PR #18, opened 2015-11-13
Add two more presentations by startup team; fix formatting

hegner:
Added links to presentations by Pere to CMS and by me to ATLAS
formated the links for the other presentations
added missing authors and dates


jouvin:
I'd suggest to put the list of presentation in reverse chronological order. As the list will grow, we'll tend to expose first the less relevant material...

What do you think?

hegner:
good point. done.

PR #19, opened 2015-11-13
Add a GitHub survival kit for beginners

jouvin:
@HEP-SF/startup This is an attempt to produce a vademecum for GitHub/Git beginners and lower to barrier for contribution. I tried to keep it as simple as possible, still explaining the few things that reduce the magics...

As it is right now, it must not be committed. I have a problem with the command examples: the syntax I am used to use in other projects doesn't work here (line breaks are lost). Probably coming from the fact that the file was edited on Windows. I'll fix it asap but this may allow an early review of the main contents.

### DIFF
--- a/github-beginners.md
+++ b/github-beginners.md
@@ -1,0 +1,119 @@
+---
+title: GitHub for Beginners
+author: Michel Jouvin
+layout: default
+---
+
+This pages is a GitHub and Git survival kit for people not familiar with these tools! This is not a documentation or tutorial about these tools. Many excellent ones are already available, including:
+
+* [GitHub Bootcamp](https://help.github.com/categories/bootcamp/), in particular [fork a repository](https://help.github.com/articles/fork-a-repo)
+* [Git for the Novice](http://swcarpentry.github.io/git-novice/) from [Software Carpentry](http://software-carpentry.org/)
+* [Git Book](https://git-scm.com/book/en/v2): the definitive Git reference! Certainly not for beginners!
+
+
+## GitHUb Workflow
+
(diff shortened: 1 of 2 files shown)


hegner:
Hi @jouvin ! I don't think we should have any github tutorial of that kind on our website. It should go better onto wikiToLearn or alike as it is of general use.- if there is anyhow a need to write something ourselves.

jouvin:
@hegner I'd say that the first reactions I have seen when we switched the web sites led me to conclude the opposite... Clearly many of us are not comfortable with the GitHub workflow and my experience is that with a little bit of explanation, it makes the buy-in easier/greater. Else we are at risk of being forced to give write access to the project repo to everybody who wants to contribute so that they can do it through the web directly in the project repo. For me this is the worst that can happen...
You are right that we should certainly not write a Git or GitHub tutorial. I tried to avoid doing this. But I think we need to provide a very minimal step-by-step startup doc so that people can start to contribute without too much pain and without doing/reading a lot of tutorials. With the hope that they will ike the tools enough to read them at some point!
As for the suggestion to put it in WikiToLearn, I have mixed feelings. WikiToLearn is really for tutorial and clearly I am not volunteering to do one.

jouvin:
Windows end-of-line issue fixed: verbatim sections should render correctly now...

hegner:
Hi Michel, are you fine w/ your changes now? Should we restart reviewing?

jouvin:
Yes, for me the contents was ready to be reviewed since the beginning... And as it is now, the formatting problem is solved and the PR can be merged if its contents is considered appropriate. Else I'll be happy to address remarks...

hegner:
Could you remove the [WIP] if you are happy w/ your current setup? Thanks.

PR #50, opened 2016-05-25
Update packaging section and make language corrections

sextonkennedy:


jouvin:
Thanks for the polishing!

PR #70, opened 2017-01-03
Main page: display events in chronological order

jouvin:
Currently, if several "breaking news" are displayed on the main page, they are in reverse chronological order (the first event to occur being last in the list). This is not very much expected, IMO. With this PR, events are displayed in the order in which they occur.


hegner:
@jouvin - I agree. Your order makes more sense.

PR #72, opened 2017-01-25
cwp: Clarifiy how to register to the CWP Google groups

jouvin:
Clarify how to register without a Google account.


peremato:
I am merging the request since nobody reacted

PR #78, opened 2017-02-01
Cleanup GSoC proposal

hegner:
Cleanup of last GSoC proposals. Using categories for filtering instead of posts

PR #79, opened 2017-02-01
Addding a page about reviews

hegner:
includes PR #78


jouvin:
@hegner : it should be rebased on upstream/master. As it is it contains, GSoC cleanup... (you need 2 different source branches to have them in //).

hegner:
@jouvin - I know. It is in the comment actually ;-)

jouvin:
But this is not a very good practice...!!! In my opinion at least...

PR #76, opened 2017-01-28
Configure Travis to run htmlproofer on PRs

jouvin:
`htmlproofer` is a Ruby module to check consistency of a web site. In particular it detects incorrect construct and dead link. It seems important to run it to ensure the quality of our website.

Once the PR has been merged, if you want to check on your machine that `htmlproofer`validates your changes, you need to execute:
```
./travis-build
```
But you can just wait for Travis to run it after you pushed something...

This PR is a second attempt, after the closed #56. It disabled checks of external links (too many failings): this is may be something to review in the future.

This PR is ready for merging (as sson as tests are ok!) but may require rebasing after #74 is merged (I suggest merging #74 first as it is probably easier and less urgent to rebase this one)...

PR #81, opened 2017-02-01
GSoC TMVA proposal: fix mentor emails

jouvin:
TMVA-related GSoC proposals list a few mentors with their emails. This has two problems:
* [ ] A major one: the pages don't validate (see #76) because the email is invalid. To avoid spamming the address has been written `AT.NOT.SPAM.` and is then invalid. I think we need to produce valid pages (that can be checked successfully with a validator like `htmlproofer`) and I suggest to either put the real emails or to write the email as it is now but not declare it as a `mailto` (but you will look the automatic filling of the subject). I have not done anything else to fix this, waiting for your input.
* [x] A minor one: `mailto` url has been written using HTML rather than markdown. I think this is better to stick to MD as much as we can. This one is already fixed. 

Adding @omazapa who is the original author of the pages.

### DIFF
--- a/_gsocproposals/proposal_TMVAadditional.md
+++ b/_gsocproposals/proposal_TMVAadditional.md
@@ -17,11 +17,11 @@ Toolkit for Multivariate Analysis (TMVA) is a multi-purpose machine learning too
 Strong C++ skills, good understanding of machine learning algorithms
 
 ## Mentors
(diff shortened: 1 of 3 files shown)


jouvin:
In fact, a compromise to have the email considered valid by an html checker but invalid for a spammer is to replace `-AT-` by `@` but keep the `-dot-`. Personnally this is not my preferred solution as people who will click on the link will probably not notice the `-dot-` and send the mail to an unknown user (and if this is me, miss the error report!).

PR #77, opened 2017-02-01
GSoC proposals for TMVA

omazapa:
added .md files with TMVA GSoC projects.

Best,
Omar.

PR #82, opened 2017-02-02
Added transfer project

vkuznet:
Please take new CMS project for Gsoc

### DIFF
--- a/_gsocproposals/proposal_TRANSFER.md
+++ b/_gsocproposals/proposal_TRANSFER.md
@@ -2,7 +2,7 @@
 title: Next generation of CMS data replication system
 layout: plain
 project: TRANSFER
-organization: CERN
+organization: Cornell, Univ. of Bolognia
 ---
 
 # Description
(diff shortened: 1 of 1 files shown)


_gsocproposals/proposal_TRANSFER.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Next generation of CMS data replication system
+layout: plain
+project: TRANSFER
+organization: Cornell, USA and Univ. of Bologna, Italy
```
hegner: can you change that to CMS? And I will add CMS to the participating organizations on http://hepsoftwarefoundation.org/activities/gsoc.html

jouvin:
I also suggest that we merge #76 first to avoid the problem we had with #77.

_gsocproposals/proposal_TRANSFER.md
vkuznet: @hegner, done, I put CMS as organization, the univ were requested by Sergei though.

hegner:
Has been handled with #84

PR #101, opened 2017-02-03
GSoC: refactoring of project descriptions for easier management

jouvin:
As it was done for organization, use a common liquid template that produced the actual project page, with all the project description done in the front-matter section of the project MD file. This should be simpler for users and less error prone (no Liquid code copy/edited with the risk of a mistake, quite easy with this unusual syntax).


hegner:
@jouvin  - I think you are overdoing it now. Don't want to expose others to much more than just providing some markdown to us. In addition, we shouldn't make the way how people submit proposals a moving target. Can do improvements for next year.

htmlproofer.log
```diff
@@ -0,0 +1,555 @@
+Running ["ScriptCheck", "LinkCheck", "ImageCheck", "HtmlCheck"] on ["./_site"] on *.html...
```
hegner: you don't want that in!

Gemfile.lock.ruby-2.0
```diff
@@ -0,0 +1,146 @@
+GEM
```
hegner: you don't want that file in!

hegner:
@jouvin - I could live with a new layout instead of user-added code.

jouvin:
As you want but I think it will quicly be a nightmare to manage and changing when there will be more proposals/projects will be more difficult. My change is in fact a huge simplification I think : to create a project (not a proposal, I don't intend to change anything there), you just produce a 2 or 3 lines front-matter. Less risk of things going wrong. And the description field, apart from being idented is pure markdown.

hegner:
@jouvin - we can do that for next year... so let's keep it WIP and tackle that later.

jouvin:
The doc changes are very limited. This is just a clarification project vs. proposal (in fact I misused project in the my last mods...

htmlproofer.log
jouvin: Sorry for the mistake... fixed.

Gemfile.lock.ruby-2.0
jouvin: idem

jouvin:
@hegner this up to you to decide but as for me this is ready for merging and this should really make simpler adding projects (not  proposals), still keeping a common structure. It was not a huge work after toine the work for the organization (this is a minor variant).

As for me this is no longer WIP in the sense that this is complete. I'd prefered to see it merged now as every addition of a project or organization (and I think we can expect quite a number of new ones) is creating a conflict that must be handled and this is wasted effort IMO. I don't plan other changes for this year!! (even though a rather trivial one along the same lines would be to turn projects and organizations into collections and generate the list one GSoC main page... but no time for this year!).

jouvin:
I think #106 demonstrates one of the benefit of this change (projet and organization name not sensitive, avoiding the risk of different spelling in project and proposals when the name is a mix of lower/upper case). If you finally agree to merge it, I suggest merging all the pending PR with new projects (#104, #105 in particular first) and I will take care of the necessary changes and push them in this PR.

hegner:
I think you are right. I was confused by your mixup of project and proposal. Before merging #104 I'd need to add Pythia and MIT to the front page. Will do that later today.

hegner:
@jouvin - feel free to start now

jouvin:
@hegner should be ok now and ready for merging.

PR #104, opened 2017-02-04
Added MIT Pythia proposal

mityinzer:


_gsocproposals/proposal_Pythia.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Expected results**:
+
+Working implementation of electron-ion collisions in Pythia 8.
+
+**Requirements**: C++
+
+**Mentors**:
+
+  * Mike Williams  mwill@mit.edu
```
jouvin: Any good reason not to use the standard syntax for emails `[name](mailto:email)`?

mityinzer: I copied the syntax from an existing proposal file (I think it was the ROOTmpi one). I'm fine with changing it if most are using syntax that (presumably) auto-generators a mailto link.

PR #106, opened 2017-02-04
Fix SciKit-HEP name so that project is properly classified

etejedor:


jouvin:
@hegner A problem that would not happen if #101 was merged! (as project names and organization names are not case sensitive in #101 implementation).

PR #123, opened 2017-02-08
Add gsoc proposal for IT/Kubernetes/CephFS

rochaporto:


etejedor:
Hi Ricardo,
We might need to change the project name, since "IT" is more a department of one of the organizations involved.
Are you contributing to Kubernetes with this driver? Perhaps we could use Kubernetes as project?

rochaporto:
Updated changing project name to Kubernetes.

PR #130, opened 2017-02-28
[WIP] Fix personal data and affiliation

xvallspl:


jouvin:
I think we should not merge a PR in WIP status without changing the title... else it makes the follow-up more difficult... If this is in WIP, probably there is a good reason and it's worth mentioning it has been fixed!

xvallspl:
Hi,

Just to clarify: this was marked as WIP because there is another commit coming with the details of my institution once I get the approval from their communication department (or enough time passes).

But I can also add those changes in an independent PR.

PR #131, opened 2017-02-28
Add UJI as an institution

xvallspl:


xvallspl:
Hi, don't know why the build fails.

I build it locally and it does generate and find the UJI.html page it's complaining about here...

jouvin:
@xvallspl you created a file `uji.html` but referenced it as `UJI.html`. This is case sensitive. May be you ran the test on Windows which ignores casing of filenames...

PR #136, opened 2017-03-30
A new page for the Visualization WG

ric-bianchi:
I have created a new page for the Visualization WG, to have a place where we can collect documents, links to the meetings, and so forth.

Beside that, I fixed some typos in the GitHub tutorial.

Have a good day,

    Ric.

### DIFF
--- a/_activities/visualization.md
+++ b/_activities/visualization.md
@@ -1,0 +1,82 @@
+---
+title: "Visualization"
+author: "Riccardo Maria Bianchi"
+layout: default
+---
+
+# Visualization
+
+The Visualization Working Group (WG) gathers experts, develeopers, users and all people from the HEP community interested in interactive data visualization and event displays. The WG has been started in October 2016 and it is convened by Riccardo Maria Bianchi (*University of Pittsburgh*). 
+
+All are welcome to join the WG and participate!
+
+## Goal
+
+The goal of the WG is to discover common approaches, foster collaboration among HEP experiments and groups, plan future software development and draft the guidelines for the future HEP visualization tools, with a view of 5-10 years.
+
+----
+
+## Description of the WG
+
+Here follows the description of the WG, its challenges and the questions which it should address, as defined by the WG itself during the Visualization session in the HSF Workshop in San Diego (see link in the meetings' list below).
+
+The first task of the WG is the planning and the writing of the [CWP](http://hepsoftwarefoundation.org/activities/cwp.html) chapter on Visualization.
+
+### Scope
+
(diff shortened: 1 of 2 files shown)


jouvin:
@ric-bianchi thanks for your contributions! But your `visualization.md` is missing a *front-matter*  section. Look at other files in the same directory (ignore `redirect_from` from existing files). Without it your page will not show up in the menu...

There is also a test error due to transient errors in the external link checks... Let's see if it disappeared, once you fixed the problem in the page.

ric-bianchi:
Thanks Michel! You are right, I will add the front matter section and I will commit a new version.

Thanks,

    Ric.

ric-bianchi:
Done, I now make a new request

hegner:
@ric-bianchi - we can just continue here.

jouvin:
@ric-bianchi Why a new pull request? pull requests are intended to be updated by new commits or by rebasing the source branch...

jouvin:
@ric-bianchi ideally you could sqash all your `visualization.md` commits into one, using `git rebase -i` (but keep the gitub-beginners in a separate commit)... but if you prefer not to do it, just tell it! (if you do it, you will have to `git push -f` to update the PR).

About the content, it seems very much related to the CWP WG. Is it the same or is it something that you would like to make more permanent than the CWP. If this is the case, it may be good to tell it a little bit more explicitly in the text...

ric-bianchi:
Hi Michel, Benedikt, 

I have one question about the layouts:  I saw some pages uses the "default" layout, while some others use the "plain" one. Is there one which is preferred?

Thanks,

    Ric.

jouvin:
default...

ric-bianchi:
Ok, I will change the layout then.

Concerning the page content, ideally it is a page where the WG can collect all the relevant material, links to external pages and tools, and the links to all the meetings, concerning the CWP or not.

Moreover, I would not bloat the CWP page with all material from each working group.

What do you think?

jouvin:
This is fine with me! And I find it a very good initiative! Reading the page I was not completely sure and I'd in mind it could be stated a little bit more clearly than it was an activity encompassing the CWP work but larger...

ric-bianchi:
Thanks Michel,

I have just added a sentence about the CWP, as the first task of the WG.

jouvin:
Can you sqash your commits? This should be easy to do with something like:
```
git rebase -i HEAD~9
git push -f
```
During `git rebase -i`, you will have the opportunity to reorder your commits (moving the lines). For example put your fixes to githb-beginners as the first of your commits. And then change to 'f' rather than 'pick' any line after 'Create virtualization...'.

ric-bianchi:
Thanks, Michel! :)

And have a nice day!

PR #143, opened 2017-05-30
Add Annecy workshop announcement and event

jouvin:


pelmer:
"many topical working groups have been formed to discuss one particular aspect of the CWP" might be better as "many topical working groups have been formed to discuss the various aspects of the CWP"

jouvin:
Fixed.

jouvin:
The error is due to a problem with Springer site (related to the Journal on SW and Computing for Big Science). I contacted the journal, I propose to ignore the error for now.

jouvin:
After checking with my contact at Springer, there is a known issue with their web site. I suggest to ignore this error for the time being... Merging this PR as we need the announcement out asap...

PR #145, opened 2017-07-20
Add HSF 2017 workshop summary talk

hegner:


jouvin:
@hegner we are suffering from all the people who push things directly (or edit on the web) and then don't go through the tests... The problems appear when there are unrelated mods. It seems we have another US web site that has a problem with the htmpl-proofer:  http://science.energy.gov. Please, could you add it to .travis-scripts/url-ignore?

hegner:
@jouvin - done

PR #157, opened 2017-09-11
[WIP] Version of htmlproofer that ignores CA errors

graeme-a-stewart:
New version of htmlproofer that adds the option
--http-ca-ignore and thus suppresses certificate
problems in the Travis CI

Patch to our private html-proofer script that
uses this special htmlproofer version


jouvin:
I let @hegner have a look and merge the PR if he agrees.

graeme-a-stewart:
But wait, there is more to come (WIP!).

Yes, I will propose the change to `htmlproofer` upsteam.

graeme-a-stewart:
Damn it - I still get lots of errors from html-proofer when running in Travis that I don't see locally. Do you know any way to replicate a Travis CI environment stand alone?

jouvin:
I remember having the same problems in my tests. I think that Ruby is not using the standard system certificates but has its own version of the supported CAs and this is probably the problem. This is strange that disabling certificate check is not enough to work around the problems.

graeme-a-stewart:
I think it's a different error from the CA one. It's a connection error, giving a status code of 0, which means, I think, Travis got no response at all from the remote site.

The error `301 Peer certificate cannot be authenticated with given CA certificates` has gone, so the patch has done some good.

Is Travis just having connection issues..? I wonder if I can tweak the timeout.

jouvin:
I also found that Travis had sometimes connection issues: it was clear as rerunning the test in Travis was not giving the same error. Not sure what can be done. This is a bit strange because I have not seen the same problem in other GitHub projects where we use html-proofer... You are right that adjusting the timeout may help.

And there are a couple of sites where clearly html-proofer fails (locally or in Travis), because robots are forbidden access...

I'd say that if this PR already fixes the certificate verification problem, I'd be in favor of merging it. It is already an improvement and doesn't prevent opening a new one for the other problem.

graeme-a-stewart:
Hi @jouvin. I looked up the connection timeout parameters for html-proofer and they are set quite short: 

https://github.com/gjtorikian/html-proofer/blob/master/lib/html-proofer/configuration.rb#L38

I am trying a patch that increases them...

graeme-a-stewart:
OK, just to record some progress:

In the last patch I enabled a `--max-connections` option that can change the number of concurrent connections from the default of 50. However, even reducing this to a very low number doesn't help. I basically concluded then that the errors generally have nothing to do with timeouts or connection concurrency.

My next tack was to try and find a way to produce locally the Travis environment to be able to debug with shell access.

1. Clone the Travis ruby container: `docker pull quay.io/travisci/travis-ruby`
2. Start the container and change to the `travis` user, then replicate the Travis ruby environment:
```
$ docker run --volume $(PWD):/srv/jekyll --interactive --tty quay.io/travisci/travis-ruby:latest /bin/bash
root@0c282bccb343:/# su - travis
travis@880a14ccb868:~$ export NOKOGIRI_USE_SYSTEM_LIBRARIES=true
travis@880a14ccb868:~$ mkdir -p ~/.gradle && echo "org.gradle.daemon=false" >> ~/.gradle/gradle.properties
travis@880a14ccb868:~$ rvm use 2.2.7 --install --binary --fuzzy
...
travis@880a14ccb868:~$ cd /srv/jekyll/
travis@880a14ccb868:/srv/jekyll$ bundler install
...
travis@880a14ccb868:/srv/jekyll$ ./.travis-scripts/html-proofer
...
```

(Note that in the `rvm` command when I asked for Ruby `2.2` I got `2.2.0`, which doesn't work as the `activesupport` gem requires Ruby `>= 2.2.2`.)

Now running the proofer I do not get the same errors as Travis, but I get a very similar error that is reproducible from the command line:

```
- ./_site/jekyll-beginners.html
  *  External link https://www.ruby-lang.org/en/downloads failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: SSL connect error
travis@880a14ccb868:/srv/jekyll$ curl -I https://www.ruby-lang.org/en/downloads/
curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

Reading around it seems this comes from a failure to negotiate an appropriate protocol and cypher between the server and client. Most likely the default tried by curl on trusty is too old for the more securely configured sites. This is also why such URLs always work just fine from our more up to date machines, which have a different set of SSL defaults. It may be possible to work around this with a suitable curl configuration or environment variable setting, but this seems rather poorly documented and I haven't found out how to do it yet.

jouvin:
Thanks for all this work...

graeme-a-stewart:
OK, I think I have a fix for these underlying SSL errors. I compiled `openssl-1.0.2l` (upgrading from `1.0.1f`) and then used `LD_LIBRARY_PATH` to use the new object files and things work. (I did also try various SSL/TLS options directly in curl, but none of them helped.)

I also found a way to update the CA certificates on trusty, using https://whatsmychaincert.com/, so in fact what I now propose is:

- Add the new openssl `.so` into git (it's 3MB, but building it very time sems silly)
- Add proper certificates for the websites that can't be validated with the default chains
- Tweak `html-proofer` to set `LD_LIBRARY_PATH` for the site validation and add the new certificates

This means that the original intention to patch `htmlproofer` to ignore CA errors starts to look moot. I think this new solution is better.

graeme-a-stewart:
There is now a new PR for the updated fix: https://github.com/HEP-SF/hep-sf.github.io/pull/160

PR #160, opened 2017-09-18
[WIP] Use updated SSL libraries

graeme-a-stewart:
Use some updated SSL libraries, compiled from 1.0.2l, to overcome observed SSL connection errors when validating some https websites.


graeme-a-stewart:
- Using the new SSL libraries is working
- Update of the certificates within the image is also working, however, it requires changing from a container to a VM (`sudo: required`), which makes the CI more resource heavy and slower

Despite these changes (which work) there are still strange errors appearing in htmlproofer that now do not seem to be related to underlying curl/ssl errors - I put some direct curl calls into the script to check that (e.g., https://travis-ci.org/HEP-SF/hep-sf.github.io/builds/276898283).

So still some more work to understand those problems.

jouvin:
@graeme-a-stewart is the last errors you are mentioning purely random or are they affecting always the same (limited) set of URLs? If this is always the same URLs (but I'm afraid this is not the case), I'd suggest to add them in the ignore list for the time being and merging what we have so far as it should already been a huge improvement...

graeme-a-stewart:
Hi @jouvin - it's very consistently the same URLs that fail (e.g., https://conda.io/docs/), but these don't fail when directly accessed by curl, which is puzzling. A tidied up version of this PR should probably be taken to at least get the solved problems solved. I can prepare that tomorrow.

jouvin:
If this is always the same URLs, I propose that for the time being, this PR includes them in the ignore list so that html_proofer can succeeds. After merging this first PR, we can try to follow up on the remaining issues...

graeme-a-stewart:
Damn it! So I tidied up this branch, checked everything, then did a final commit to ignore all of the URLs with the mysterious failures (adding the `data-proofer-ignore` tag). **But they all still fail!**

Somehow they must be failing internally in `html-proofer`, which at least does explain why the error was not seen directly with `curl`. This is so frustrating!

hegner:
At this point in time I would just go for a docker image in travis... as it is now things are not debuggable and every iteration is costly

jouvin:
@graeme-a-stewart You should use the file `.travis-scripts/url-ignore` to define the URL to ignore. I suspect that the way the script sets this option override the option you defined. An it is much more flexible normally...

graeme-a-stewart:
OK, let's try again in https://github.com/HEP-SF/hep-sf.github.io/pull/162

PR #163, opened 2017-09-26
make links more visible

hegner:
Pere reported that colorblind people can hardly spot hyperlinks. Making them slightly brighter now.

### DIFF
--- a/css/hsf.css
+++ b/css/hsf.css
@@ -188,7 +188,7 @@
 }
 
 a {
-  color: #dd0000;
+  color: #cc5555;
 }
 
 a:hover {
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
This makes the links rather bright red (obviously), I wonder if 0xcc3333 is better - light, but a less vivid red. It's probably a good idea to have links underlined too.

css/hsf.css
```diff
@@ -188,7 +188,7 @@
 }
 
 a {
-  color: #bb0000;
+  color: #cc5555;
```
graeme-a-stewart: After some discussion with @peremato can you change to `#cc5555`.

graeme-a-stewart:
Travis failures do seem to be reflecting some website issues, but I think they are temporary, so I'll merge this.

PR #166, opened 2017-10-09
Add short HOWTO for editing the community calendar

graeme-a-stewart:
This PR adds a short section on how to edit the community calendar, publicising the URL that you need to use to add the HSF calendar to your own calendar application. Instructions are given for Google Calendar.

I tried the URL with a throw-away Google account that I have and it would not let me add events, so the editing permissions do seem to be locked down (@pelmer can probably confirm it's the startup team).

I added this link under the Communication navigation item.

### DIFF
--- a/_includes/navbar.ext
+++ b/_includes/navbar.ext
@@ -31,6 +31,7 @@
                   <li><a href="/technical_notes.html">Technical Notes</a></li>
                   <li><a href="/newsletter.html">Newsletter</a></li>
                   <li><a href="/organization/minutes.html">Meeting Notes</a></li>
+                  <li><a href="/calendar.html">Community Calendar Editing</a></li>
                   <li class="divider"></li>
                   <li><a href="http://www.hepsoftware.org">HEP S&amp;C Knowledge Base</a></li>
                 </ul>
--- a/_includes/navbar.ext
+++ b/_includes/navbar.ext
@@ -31,7 +31,6 @@
                   <li><a href="/technical_notes.html">Technical Notes</a></li>
                   <li><a href="/newsletter.html">Newsletter</a></li>
                   <li><a href="/organization/minutes.html">Meeting Notes</a></li>
-                  <li><a href="/calendar.html">Community Calendar Editing</a></li>
                   <li class="divider"></li>
                   <li><a href="http://www.hepsoftware.org">HEP S&amp;C Knowledge Base</a></li>
                 </ul>
--- a/calendar.md
+++ b/calendar.md
@@ -4,9 +4,9 @@ author: "Graeme Stewart"
 layout: default
 ---
 
-The HSF [Community Calendar](https://calendar.google.com/calendar/embed?src=e4v33e1a1drbncdle1n03ahpcs%40group.calendar.google.com&ctz=Europe/Amsterdam)
+The HSF [Community Calendar](https://calendar.google.com/calendar/embed?src=e4v33e1a1drbncdle1n03ahpcs%40group.calendar.google.com){:target="_hsf_calendar"}
 can be edited by people in the [HSF Startup Team](https://groups.google.com/forum/#!forum/hep-sf-startup-team)
-Google Group.
+Google Group **TBD** an additional group **TBD**.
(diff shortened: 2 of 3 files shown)


graeme-a-stewart:
I was actually able to check the editing permissions - it's currently 

- hepsoftware@gmail.com
- hep-sf-startup-team@googlegroups.com
- peter.elmer@cern.ch

which seems just fine to me.

_includes/navbar.ext
```diff
@@ -31,6 +31,7 @@
                   <li><a href="/technical_notes.html">Technical Notes</a></li>
                   <li><a href="/newsletter.html">Newsletter</a></li>
                   <li><a href="/organization/minutes.html">Meeting Notes</a></li>
+                  <li><a href="/calendar.html">Community Calendar Editing</a></li>
```
jouvin: Why not to add this menu  to 'Get  Involved' rather than 'Communication'?

calendar.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Community Calendar - Editing"
+author: "Graeme Stewart"
+layout: default
+---
+
+The HSF [Community Calendar](https://calendar.google.com/calendar/embed?src=e4v33e1a1drbncdle1n03ahpcs%40group.calendar.google.com&ctz=Europe/Amsterdam)
```
jouvin: May be remove the timezone in the URL?

jouvin: And after the `(link)`, may be add `{:target="_hsf_calendar"}` (no space between `()` and `{}`) to get the calendar in a new tab, being able to continue to read the doc...

calendar.md
```diff
@@ -0,0 +1,29 @@
...
+can be edited by people in the [HSF Startup Team](https://groups.google.com/forum/#!forum/hep-sf-startup-team)
```
jouvin: I still don't like the idea that we use one group for 2 different purposes. Why not to create (it is so easy) a group `hep-sf-calendar-editors` where we could add people who may be legitimate editors and don't want to be in the startup team list (despite the notification trick). We can then just add this group to the calendar sharing list (any startup team member can do it) and document the usual `+subscribe` trick to subscribe to this group. Additionally this will give us the ability to contact calendard editors if we need to share some information, something not possible if people disabled email notifications...

calendar.md
```diff
@@ -0,0 +1,29 @@
...
+
+```
+https://calendar.google.com/calendar/ical/e4v33e1a1drbncdle1n03ahpcs%40group.calendar.google.com/public/basic.ics
+```
+
+Then you will have the ability to view events in the community 
+calendar and, if you are in the startup
+team group, add them. 
+
+Please ask to join the startup
```
jouvin: See my previous comment... To be adjusted if we go for another group.

_includes/navbar.ext
graeme-a-stewart: I'm ok with that. As *Get Involved* is a single page link, do you think just add calendar editing as part of that page? It's short enough that a separate page is perhaps unnecessary.

calendar.md
graeme-a-stewart: Fine be me, done for the next commit,

calendar.md
graeme-a-stewart: I can't say I deeply care, but if we do this I think the startup team should be able to also edit (I suppose you were proposing that, right). Then the new group is for extra editors (and should be approved subscriptions I think).

graeme-a-stewart:
@jouvin I pushed some updates in line with your suggestions, but I made this a WIP pending the decision on the second editing group. Let's see what others think - for me it's fine either way.

jouvin:
@graeme-a-stewart I'm fine with your changes. Let's wait the decision on the editor group to finalize it, as you suggested.

graeme-a-stewart:
@jouvin - should be good to go now

jouvin:
LGTM! BTW, have you added the editors group to the calendar with the appropriate rights (probably adding/modifying events is enough, startup team has the right to manage the sharing if I'm right).

PR #167, opened 2017-10-11
Fix messed up bullet list in CWP page

jouvin:
I'd suggest that we use the workflow we recommend (PR) for every change to the web site, even the trivial ones! This gives a chance to others to spot a trivial mistake that can happen to everybody...

### DIFF
--- a/_activities/cwp.md
+++ b/_activities/cwp.md
@@ -57,7 +57,7 @@ we are reaching out to other HEP experiments around the world to participate.
     * Draft WG white paper: [google doc](https://docs.google.com/document/d/1_YKGs8waBidpS8akmjxEVjV00DThINbJAuVelzk2uHg/edit) - **available for review**
   * Data Analysis and Interpretation WG
     * Draft WG white paper: [google doc](https://docs.google.com/document/d/13ediEyaP1QGF-esZOXUH3SNShJFsqa2f6hVmEukJv_c/edit) - **available for review**
-* Data and Software Preservation WG
+  * Data and Software Preservation WG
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Damn it - it's my bad for trying to push in some quick fixes in some of the URLs.

Could we not also push a preview version of the website, so we can validate the output of the changes and not just rely on inspection of the markdown?

jouvin:
@graeme-a-stewart I don't think that pushing a preview version really makes sense (or at least this is difficult to implement). But http://hepsoftwarefoundation.org/jekyll-beginners.html documents how to easily install Jekyll on your machine so that you can check the effect of any change before pushing it... This is what I'm doing personnally: I would not be able to spot most of the things I find just looking at markdown with my old eyes...

graeme-a-stewart:
The preview isn't for the committer (yes, I also run a jekyll locally), it's for the reviewer to check nothing was missed/messed. We do do this for the ATLAS software documentation site and it helps a lot (though that means it's probably not directly liftable from out gitlab implementation).

jouvin:
I don't really know how to do it... but I think that all reviewers of the web site changes should have their Jekyll instance configured on their local machine and that it is probably enough to solve the pb. At least this is my personnal practice: I always checkout the source branch of the PR and run Jekyll out of it to see the result for the modified pages. In general, before looking at the markdown...

graeme-a-stewart:
I have to confess that I haven't been doing that so far - so I should start.

BTW, is there a way to actually forbid direct editing via GitHub?

jouvin:
Yes this is possible! This is what [protected branch](https://help.github.com/articles/about-protected-branches/) is designed for. As it is possible to require a review on these branches.

PR #171, opened 2017-10-19
Update to new Data Preservation document

graeme-a-stewart:
The Data Preservation link was wrong - it was just to the notebook


hegner:
The commit message is a bit misleading (compared to the message in this pull request), but OK

PR #180, opened 2017-11-08
Page for link to talks and archived CWP documents

graeme-a-stewart:
As discussed in several weekly meetings, it's useful to archive CWP chapters, especially
for places where Google Docs are blocked.

I also wanted to collect CWP Roadmap talks and keep them off the main page to stop
it growing forever.

This I did in the same page (please say if you think it would be better to use separate pages, I
was undecided).

I also fixed the names of a few of the working groups and updated the Frameworks link
to the new summary page.


jouvin:
I have mixed feelings about putting the PDF in the repo if they are not final... This makes the repository unnecessarily big... but I don't have a real alternative to propose...

graeme-a-stewart:
Hi @jouvin - yes, I agree with you. I had forgotten we had a documents repo for just this reason. I will prepare an update for that to upload the CWP PDFs and then a new PR for the main website to link to them.

BTW, there are some ridiculously large images in the repo now as well, that should never have been allowed. *sigh*

PR #182, opened 2017-11-09
CWP snapshot documents

graeme-a-stewart:
Add new page for links to snapshot documents
Link to Roadmap paper in Documents archive (start clean up for removing large binary objects)

Make a few fixes in working group names
Remove .DS_Store OX X binary file

Note, use a Jeykll variable to avoid long URL strings


_data/hsf.yaml
```diff
@@ -0,0 +1,4 @@
+# Useful variables that avoid excessive typing and typos!
+# Refer to these in Jeykll as, e.g., {{site.data.hsf.docs}}
+docs: https://github.com/HEP-SF/documents/blob/master
```
jouvin: Why not put all these variables in `_config.yml`?

graeme-a-stewart: Because it doesn't get reloaded on changes (see the note in https://jekyllrb.com/docs/configuration/ - quite far down) and you need to restart Jekyll. If the data files change then Jekyll reloads which is more convenient and avoids confusion.

graeme-a-stewart:
The .DS_Store is in the current .gitignore, but that came after a version had been included in the repo (and it doesn't force it out, it needs manually removed).

_data/hsf.yaml
jouvin: I personnally, based on my Jekyll experience, find it quite unusual to define these variables outside `_config.yml` (except may be if you have tons of them!). It is true that Jekyll doesn't reload this file when it is modified but AFAIK, GitHub Pages does restart Jekyll if it is changed. I always saw my changes to this file immediately picked up by GH Pages without any additional actions....

graeme-a-stewart: However, when you're developing and need to change any variables it's much slower to have to stop and restart Jekyll, which is why I like these "non-core" things to be defined in this way. FYI, the ATLAS docs site uses these definitions quite a bit - having them in different "namespaces" can be handy. They are completely standard and supported part of Jekyll for defining Liquid variables (https://jekyllrb.com/docs/datafiles/).

jouvin: I know that nothing prevent them and that this is completely standard... I would not say that because ATLAS using it, it is demonstrating that this is a good practice! Anyway this is not a blocking issue, clearly.

graeme-a-stewart:
I agree this is a minor issue - I switched to using `_config.yml` instead.

PR #183, opened 2017-11-10
Weekly meeting 2017-11-09

graeme-a-stewart:


graeme-a-stewart:
You know, I think we have reached the stage with the HSF website that we have so many external links that the chances of a spurious failure are significant. I don't know how best to deal with this. :-(

hegner:
Ok. So we should maybe move to having a nightly for checking and move it out here. And check only URLs that are in the diffs

jouvin:
I'm  a little bit surprise we have so many problems... This is true we have many external links. I also see that it generally works if you do it twice... What about rerunning (immediatley) the URL check in case of failure. It may be enough to get a green test in 99,9% of the cases... I've the feeling that moving to a nightly build approach is both an overkill and something that will make more difficult to clean the real problems (you may not know what is the appropriate URL for a broken one and need the author to do it. For example for GSoC proposals...

graeme-a-stewart:
We should make this an issue and discuss further there.

PR #186, opened 2017-11-17
CWP page updated with draft 2 information

jouvin:


jouvin:
Test failures are in particular due to the PDF link added not yet reachable... but anyway we should ignore the other errors if any and fix this pb later...

jouvin:
BTW, I found at least one really buggy link... fixed now. The 2 other ones can be dealt with later but there are good reasons to think that `html-proofer` is right and that we need to check them...

graeme-a-stewart:
Now merged with some updates in #187

PR #197, opened 2017-12-20
Update website for CWP Roadmap release

graeme-a-stewart:
Fairly substantial update:
- Front page was updated to link to arXiv document and new CWP "result"
  page; announcement of joint Naples workshop
- Naples workshop added as an upcoming event
- Main CWP page has been rewritten to be a summary of the CWP outcome
- Old CWP page has been moved to a history page, rewritten to be in the past
  tense and only relevant information kept

### DIFF
--- a/_activities/cwp.md
+++ b/_activities/cwp.md
@@ -1,181 +1,58 @@
 ---
 title: "Community White Paper"
-author: "Peter Elmer"
+author: "Graeme Stewart"
 layout: default
 redirect_from:
   - /cwp.html
   - /workinggroups/2016/07/06/communitywhitepaper.html
 ---
 
-# Community White Paper (CWP) 
+# A Roadmap for HEP Software and Computing R&D for the 2020s
 
-## A Roadmap for HEP Software and Computing R&D for the 2020s
+In 2017 the HEP Software Foundation produced a 
+[roadmap white paper](https://arxiv.org/abs/1712.06982){:target="_cwp_roadmap_arxiv"} on the
+software and computing challenges that will be faced during the next decade.
 
-Realizing the physics programs of the planned and/or upgraded HEP experiments 
-over the next 10 years will require the HEP community to address a number of 
-challenges in the area of software and computing. It is expected that
-the computing models will need to evolve and a significant "software 
-upgrade" is required to prepare.
-In order to identify and prioritize the software research and development 
-investments required, we are now beginning a community planning process. The 
-aim
-is to produce a Community White Paper 
-(CWP) which will describe the community strategy and a roadmap for 
-software and computing R&D in HEP for the 2020s. This activity is organised under 
-the umbrella of the HSF. The LHC experiments and HSF have been 
-specifically [charged by the WLCG project](/assets/CWP-Charge-HSF.pdf) and
-we are reaching out to other HEP experiments around the world to participate.
+**The CWP Roadmap can still be signed by members of the community who endorse
+it.** Please use contact the 
+[CWP Ghost Writers](mailto:hsf-cwp-ghost-writers@googlegroups.com?subject=CWP Signature) to add your name.
+We very much encourage you to do this, to show the breadth of the community's support
+for the roadmap.
 
-## Final Roadmap Paper
+## Community White Paper Reports
 
-The second draft of the CWP, the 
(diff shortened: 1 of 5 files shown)


graeme-a-stewart:
OK, there is a serious failure in the htmlproofer - I am investigating.

_layouts/main.html
```diff
@@ -12,18 +12,22 @@
...
 <hr>
 
 <!-- Important announcements from announcements/_posts -->
 <div>
-The HSF is guiding a community process to develop a consensus roadmap for HEP Software and Computing R&amp;D for the 2020s.
-More information can be found on the
-<a href="{{site.baseurl}}/activities/cwp.html">Community White Paper (CWP)</a> page on the HSF site.
+<p class="lead">
+The HSF has now submitted the 
+<a href="https://arxiv.org/abs/1712.06982">final roadmap paper</a> from the
```
jouvin: I suggest adding a target to open the arXiv site in a separate tab, as in other pages.

_activities/cwp.md
```diff
@@ -1,181 +1,58 @@
...
-(CWP) which will describe the community strategy and a roadmap for 
-software and computing R&D in HEP for the 2020s. This activity is organised under 
-the umbrella of the HSF. The LHC experiments and HSF have been 
-specifically [charged by the WLCG project](/assets/CWP-Charge-HSF.pdf) and
-we are reaching out to other HEP experiments around the world to participate.
+**The CWP Roadmap can still be signed by members of the community who endorse
+it.** Please use contact the 
+[CWP Ghost Writers](mailto:hsf-cwp-ghost-writers@googlegroups.com) to add your name.
+We very much encourage you to do this, to show the breadth of the community's support
+for the roadmap.
```
jouvin: I personnally think that we should keep Google Docs as the main signing place. This is far more manageable that emails that we can overlook and are difficult to trace if we forget to do it immediately.

graeme-a-stewart:
Problem with links identified and corrected (there was an & in one of the page titles that becomes a drop down link).

_activities/cwp.md
```diff
@@ -1,181 +1,58 @@
 ---
-title: "Community White Paper"
-author: "Peter Elmer"
+title: "A Roadmap for HEP Software and Computing R&D for the 2020s"
```
jouvin: I agree with the page title change but it would be good to keep Community White Paper as the menu entry.  I thought I added the support for a `menu` attribute but it doesn't seem the case. Easy to implement in `_includes/navbar.ext`. I'll try to open a PR later if you don't do it yourself.

_activities/cwp.md
hegner: The problem is that we do not get notifications from that doc (thus have no idea when people add themselves) and that some additional (error-prone?) post-processing is needed. In particular we already validated all current entries and with a free-floating document we may need to redo that again.

graeme-a-stewart: I also dislike giving people two ways to do things. Give them one simple way and more people will add their names. The GDoc was great for the initial flood of people, but now it's a workflow that looks more fragile than an email followed by a PR into the document. The reply to the email from the GW team is sign off that it got handled. Once a suggestion is accepted into the GDoc it becomes rather invisible without carefully checking the history.

_layouts/main.html
graeme-a-stewart: Thanks for spotting that - done!

_activities/cwp.md
graeme-a-stewart: In the first rebase I went back to the old name because the text in the navbar became too long. If you want to change it afterwards that's fine for me.

_activities/cwp.md
jouvin: Do as you want! I am in minority! I remain convinced that the workflow I proposed was less error proof! The whole point is that GDoc is commenting only, so we have a workflow to validate what is entered. And from the GDoc, producing the text files is not really error prone, because git diff shows you immediately if you made a big mistake. In this workflow, signatures received by emails should have been entered by us in the GDoc so that it remains the **only** authoritative source...

graeme-a-stewart: Also, as we made the google doc public for comments things will come through an anonymous and then we have no 'paper trail' back to the person who actually requested the change. When the google doc was authenticated for comments many people weren't able to do it and I got many personal emails, which are really easy to lose...

graeme-a-stewart: OK, then @hegner and I pledge to take responsibility! Thanks.

graeme-a-stewart:
Last commit adds a `subject` line to the `mailto:` link that allows the GW group to better control the new author requests coming by email (thanks to @hegner for the suggestion).

This PR is ready to go now.

PR #206, opened 2018-01-22
Gsoc update

graeme-a-stewart:
This PR improves on https://github.com/HEP-SF/hep-sf.github.io/pull/205 

- Improved English and page layout for main GSoC page
- Rephrased 2017 GSoC page to make its archive status clearer


graeme-a-stewart:
@etejedor - please also check!

gsoc/2017/index.md
```diff
@@ -10,18 +10,21 @@ redirect_from: /gsoc/gsoc2017.html
...
 
-The CERN software for experiments (CERN-SFT) group has participated in the GSoC since 2011. This year the program expands to involve the high-energy physics community under the umbrella of the HEP Software Foundation.
+The CERN software for experiments (CERN-SFT) group has participated in the GSoC since 2011. In 2017 the program expanded to involve the high-energy physics community under the umbrella of the HEP Software Foundation.
 
-Students interested in participating in the GSoC can find project proposals [here](#Projects).
+Students interested in participating in the GSoC for 2017 were able to chose from various project proposals [here](#Projects).
 
 HSF GSoC Administrators: Sergei Gleyzer <a href="mailto:sergei@cern.ch">sergei@cern.ch</a> and Enric Tejedor Saavedra <a href="mailto:etejedor@cern.ch">etejedor@cern.ch</a>
```
etejedor: We need to add Antoine, who has joined the admin team this year:
- Antoine Perus (perus@lal.in2p3.fr)

graeme-a-stewart: But that is for the 2017 archive page, which we shouldn't change. Should we put the new coordinators all down on this activities/gsoc.html page maybe?

graeme-a-stewart:
There is a problem with the migration to gosc/2017 - some links are not being renamed properly and I have to work out why to then fix them.

gsoc/2017/index.md
etejedor: Oh I see, you are right. The new coordinators should then be in the equivalent page for 2018

graeme-a-stewart:
Applied various fixes to do with year handling and broken links. I am not 100% sure that organisations and projects should be in the per-year directories as these are rather likely to remain stable for multiple years. That will require some adjustments to the way that the pages are built, however.

This MR is probably good enough to get 2018 GSoC started though...

graeme-a-stewart:
N.B. The link failure should disappear with the PR is accepted - the problem is that redirects are generated with the official website URL, not a relative link.

gsoc/2017/index.md
```diff
@@ -0,0 +1,86 @@
...
+layout: default
+year: 2017
+redirect_from: /gsoc/gsoc2017.html
+---
+
+# ![CERN](/images/CERN-logo.jpg){:height="100px"} ![HSF](/images/hsf_logo_angled.png){:height="100px"} Google Summer of Code 2017
+
+## Introduction
+
+[Google Summer of Code](https://developers.google.com/open-source/gsoc/) is a program that allows students to contribute to development of open-source projects, mentored by participating organizations.
```
jouvin: The 3 first § should be removed as they duplicate the new GSOC introduction in `gsoc/index.md`.

jouvin:
Except my minor remark above, LGTM. I still think that an alternative implementation would have been to have one collection per year (every year, we would need to add one collection in `_config.yml`). This would have avoided to add a year in each MD file (as the collection will contain only the files of the year). And we'd need no common files between collections, like `gsoc/index.md`.... Another advantage would have been to preserve the existing URLs for 2017 (even if it is probabably not critical). But as this PR is already quite advanced, it is probably easier to merge it...

My Jekyll installation also suggests to rename `plugins` into `plugins_dir` in `_config.yml`:
```
Deprecation: The 'plugins' configuration option has been renamed to 'plugins_dir'. Please update your config file accordingly.
```

graeme-a-stewart:
@jouvin - should have addressed your points in the last 2 commits

graeme-a-stewart:
@jouvin - the `plugins_dir` change actually broke stuff. Reverted in https://github.com/HEP-SF/hep-sf.github.io/commit/70b90f25e496eca4ab0ff5fe853db740bb4ca12f.

PR #213, opened 2018-01-26
hot-fix rendering by year

hegner:


_gsocproposals/2018/proposal_ATLASEigen.md
```diff
@@ -2,6 +2,7 @@
 title:  Faster matrix algebra for ATLAS
 layout: gsoc_proposal
 project: ATLAS
+year: 2018
```
graeme-a-stewart: Do we need the year metadata anymore? It seems like you now get this during processing from the path.

graeme-a-stewart: If we decide we don't need it then it should be removed from the 2017 proposals and the instructions.

graeme-a-stewart:
There is a problem in the 2017/summary.md page - the skipped 2018 projects are causing a blank line in the markdown, which is then splitting the table into multiple pieces. I'll need to work on a fix for that - but we should decide if the project proposals should have year attributes in them first.

_gsocproposals/2018/proposal_ATLASEigen.md
hegner: If we fix the usage of page.year in _layouts/gsoc_proposal.html we should be able to drop the year from the metadata.

hegner:
see https://shopify.github.io/liquid/basics/whitespace/

graeme-a-stewart:
Yes, I just worked out the gymnastics. Seeing if I can do it without the `page.year` attribute...

graeme-a-stewart:
Fix for summary is in https://github.com/hegner/hep-sf.github.io/pull/7

PR #217, opened 2018-01-26
First draft of the VecCore proposal

xvallspl:


xvallspl:
oops. Thanks, Benedict!

PR #252, opened 2018-01-31
add ATLAS Rucio Metadata proposal

mlassnig:


mlassnig:
"""- ./_site/cwp-whitepapers.html
  *  External link http://science.energy.gov/~/media/hep/hepap/pdf/May-2014/FINAL_P5_Report_Interactive_060214.pdf failed: 302 Server returned nothing (no headers, no data)
htmlproofer 3.7.4 | Error:  HTML-Proofer found 1 failure!"""


one more for the htmlproofer ;-)

jouvin:
@mlassnig this is annoying... htmlproofer is doing a very good job but we have intermittent problems when it is run in the context of Travis that we cannot reproduce outside... Anyway, it is the only way to track easily trivial mistakes in URL (and in the context of GSoC, mispelled project/organization name) so we prefer to take the risk of false positive (easy to interpret).

jouvin:
BTW, restarting manually the build it often enough to get rid of the false error...

PR #254, opened 2018-01-31
Fix file ownership problem for docker container proofing

graeme-a-stewart:
~~Do not merge this - for testing only!~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,10 @@ services:
 before_install:
   - docker pull graemeastewart/hsf-jekyll
  
- # Skip install as the container does all the work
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
OK, this PR has the fix so that the new docker proofer works correctly. The trick is to `chown` the checkout so that it's owned by UID 1000, matching the jekyll user inside the docker container.

It is a mystery as to why running as root suddenly started to fail in this version of the container as the old container had been running that way without issue for a long time.

graeme-a-stewart:
I believe I found a better fix - new PR coming soon...

PR #256, opened 2018-01-31
Run container for proofing as current user

graeme-a-stewart:
This is a better fix for matching the container UID to that of the user who checked out the git repo: use docker's `--user` option to ensure the container has the same identity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 install: /bin/true
  
 script:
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
OK, there is one tiny annoyance, which is that running as an unknown user sets HOME to `/` causing a few warnings. That's now fixed in a minor update to the docker container that I just pushed to the registry.... and we're GTG!

PR #261, opened 2018-02-01
Add performance warehouse proposal

graeme-a-stewart:
Add new HSF 'generic' project
Add Giulio and Graeme to mentors list


ktf:
Weird. Thanks for spotting this, I will follow up with github. I think they changed something because it definitely used to work...

PR #278, opened 2018-02-02
Visualization project for CMS HGcal

felicepantaleo:
adding project proposal for CMS High granularity calorimeter 4d event visualization.


felicepantaleo:
@rovere fyi

PR #286, opened 2018-02-02
Minutes of weekly meeting

graeme-a-stewart:
Sorry I had no time to do these until now - would be good if they can be accepted today. Ta.


graeme-a-stewart:
Thanks @hegner!

PR #287, opened 2018-02-02
GSoC: remove data-proofer-ignore tag from Rucio proposals

jouvin:


jouvin:
It seems there is no way to get the Rucio site being validated by html proofer which is weird... Not very important but may be something to feedback to Rucio developers. It seems to come from the certificate used: it would be good to understand if they use a certificate really specific or if this is hiding a problem with htmlproofer that we'll see with other sites in the future...

Keeping this PR open as a reminder for the time being but feel free to close it if you prefer...

graeme-a-stewart:
Hi @jouvin - actually the Rucio team are aware and ATLAS Ops are trying to install a verifiable certificate to fix this. @mlassnig should keep an eye on this and let us know when we can start validating their site again.

mlassnig:
Hi, the rucio.cern.ch is an official COMODO certificate issued via the CERN https licence agreements. I don't understand why it's not being correctly certified.

graeme-a-stewart:
Checking

https://www.sslshopper.com/ssl-checker.html#hostname=https://rucio.cern.ch/

It does look much better now - it may be that the container we use for proofing lacks the COMODO certificate. I will try and fix this a la https://github.com/HEP-SF/docker/blob/master/hsf-jeykll/Dockerfile#L5

jouvin:
Thanks. It was my guess that we may just add an intermediate certificate which is missing in our container config... (with COMODO and other CAs, often there is only the root certificated in the CA bundle and you need to add the intermediate CAs used to actually deliver certificates).

graeme-a-stewart:
I have prepared an updated container that fixes this issue, but it needs #318 to be accepted first as the internal container environment changed a little.

jouvin:
@graeme-a-stewart I just accepted #318, is there anything needed to change the container before accepting this one?

jouvin:
Check needs to be restarted once the new container is in production...

graeme-a-stewart:
I tagged the new container - should be GTG...

jouvin:
And the check is now ok! So ready for merging...

graeme-a-stewart:
Merging now...

PR #313, opened 2018-02-26
New mentor and url

grasseau:


etejedor:
Hi @grasseau, could you make sure the older commits do not appear in your PR? There are 26 commits in this PR, and only the 4 most recent are the ones with new content.
Also, could you squash your four more recent commits into just one with the final result?

PR #317, opened 2018-02-27
Remove expired Google Form link

graeme-a-stewart:
The Google Form used in the logo contest seems to have expired and
is causing html check failures. Removed it and added a 'historical interest'
header.

### DIFF
--- a/logo_contest.md
+++ b/logo_contest.md
@@ -6,6 +6,11 @@ layout: default
 
 # Logo contest
 
+**Page now kept for historical interest, many thanks to Joschka Lingemann 
+for designing the winning HSF logo.** 
+
+---
+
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
The problem is it's quite a lot of work to do that and if the page is just archived (now clearly stated at the top) I think it's enough. It's actually hard to change all the language consistently as I know from doing the CWP historical pages.

It's possible we might even want to get rid of the page completely, but that should be discussed more widely.

graeme-a-stewart:
Merging this now as it fixes a real problem - happy to open an issue if you want to discuss the changes of tense with others.

eduardo-rodrigues:
That's perfectly fine I think too :-).

PR #330, opened 2018-03-21
CWP: update link to DOMA paper

jouvin:
- Publish the LaTex (Overleaf) version instead of the Google Doc


jouvin:
Error is with a CERN page ( http://openlab.cern/news/press-release-cern-openlab-tackles-ict-challenges-high-luminosity-lhc failed) which seems to have disappeared... May be we need to turn the link into text only.

PR #335, opened 2018-04-05
Packaging group minutes, 2018/4/4

graeme-a-stewart:


graeme-a-stewart:
@bencouturier thanks for your help with the notes. Things were still a bit scrappy doing a lot of taking and writing, but if you can @ktf can give a quick check it would be good. Thanks.

organization/_posts/2018-04-04-packaging.md
```diff
@@ -0,0 +1,79 @@
...
+- HSF rebranding - use the new name for the mailing list: `hsf-packaging-wg@googlegroups.com`. 
+
+
+Summary of Packaging discussion in Naples - Giulio
+==================================================
+
+- Need to now do a sensible test drive on the tools.
+
+- Test cases for the development environment probably need further expansion.
+    - Incremental builds within packages might need to be reviewed (e.g., modify ROOT and rebuild only a few libraries without recompiling every downstream piece). This is not guaranteed to work, but with *caveat emptor*, can be useful to be able to accelerate development.
```
ktf: I was actually thinking that the only case in which it does not work is when you do not type "make" in all the dependent packages. aliBuild actually does that for you, if you want, so in principle is not so bad, assuming the inner build system is sensible (e.g. configure / make or cmake) and can spot when dependencies are modified.

graeme-a-stewart: OK, that would work when you have a complete build area already. How would it work if you don't? Then you can LD_LIBRARY_PATH in your new component?

ktf: I think the requirement we have is that you have built everything at least once. Also by well behaved I mean build systems which do not need LD_LIBRARY_PATH or any other environment in order to build.

ktf: That said, aliBuild would actually perform a full build in the case you mention.

PR #342, opened 2018-04-17
Fix various broken URLs

jouvin:
Fix URLs that caused htmlproofer to fail in previous PRs.


jouvin:
Looks like the broken URLs are fixed by this PR. But it remains painful to have all the false errors: for this particular PR, I had to restart 3 times the build to have it completing successfully... I don't understand why it is so painful to get htmlproofer working properly in Travis....

PR #364, opened 2018-06-15
Remove link to broken site

graeme-a-stewart:
Download has expired, so don't try to link to it anymore (keep URL as text for "history").


graeme-a-stewart:
Still a problem with the Milan INFN site - might need the intermediate certificate for that one.

PR #373, opened 2018-07-11
Announce software forum meeting

graeme-a-stewart:
Accepting this also rebuilds the website and removes the old PyHEP and Frameworks announcements


graeme-a-stewart:
OK, all looks good - dead link was removed

organization/_posts/2016/2016-04-28-startup.md
```diff
@@ -23,7 +23,7 @@ This has worked well in some ATLAS meetings, producing more substantive meeting
 
 General agreement it is worth trying. Have to sort out how to handle the website, for which a meeting list is autogenerated from github based notes (which would disappear).
 
- - New website [proposal](https://hegner.web.cern.ch/hegner/hsf/)
+ - New website proposal: https://hegner.web.cern.ch/hegner/hsf/
```
jouvin: Why to keep the mention of a no longer existing URL. I'd rephrase the meeting summary to remove it... but can be done as a future MR to avoid delaying the merge of the meeting announcement.

graeme-a-stewart: I was taking the view that minutes are a historical record, so they should be preserved as much as possible. So I preserve where the link was, even if it's then turned inactive.

jouvin:
This is why I understood! But I'm always wondering what's the value of considering minutes as "immutable" objects if they contain reference to data that cannot be accessed anymore... A personal view though!

PR #383, opened 2018-09-21
Fix meeting navigation and move calendar

graeme-a-stewart:
Meeting navigation implemented in 3712e72ef didn't work in practice as
GitHub pages doesn't allow Jeykll plugins. This rework uses the builtin
`page.{next,previous}` navigation instead. This has the drawback that
all posts (including announcements) are part of the navigation,
instead of just meetings. However, it's a substantial improvement
on having no navigation at all.

The community calendar was moved from the "activities" dropdown
to "communication", as discussed and approved in the meeting
of 2018-09-20.


jouvin:
@graeme-a-stewart thanks for this PR, LGTM. We still have the failing tests because of IOP being very long to respond (I checked manually). I'd suggest to we disable the check on this site as I think it is important to have green tests. Please, could you do it as part of this PR?

jouvin:
For the record, second check error is about EP site at CERN which is currently under maintenance. Should not last...

graeme-a-stewart:
Thanks @jouvin! The number of sites that cause sporadic failures is extremely annoying, but as the IOP one comes up rather a lot I agree that we should skip over it (done in the last commit).

PR #393, opened 2018-10-19
Weekly meeting minutes, 2018-10-18

graeme-a-stewart:


organization/_posts/2018/2018-10-18-coordination.md
```diff
@@ -0,0 +1,156 @@
...
+    of activities.
+
+Technical Notes
+---------------
+-   Eduardo and Andy McNab following up on DOIs for some of the
+    technical notes.
+
+CWP
+===
+-   ### General Matters and Roadmap
```
jouvin: Remove the bullet marker on this line. And see my general remark about mixing the heading markup styles.

graeme-a-stewart: Sure, no problem, but AFAICS this is the markdown we have used for ages in this section, e.g., from

https://hepsoftwarefoundation.org/organization/2018/10/04/coordination.html

```
CWP
===
-   ### General Matters and Roadmap
    -   [*CSBS review
        comments*](https://docs.google.com/document/d/16T2RRu1LmAyXTgtKjyWgOwZR8zsVOw2Y1SCntot3_NU/edit?usp=sharing)
    -   Michel was too occupied to make progress on these yet (seems
        everyone else was too).
-   ### Publication status for Individual WG Papers
    -   **Software Trigger and Event Reconstruction**
        -   PR to the documents repo was
```

Changing from `===` to `#` will change nothing in the CWP title (I did try it). The current notation is what comes out from pandoc so I really don't think we want to change it without good cause.

N.B. In my Chrome and Firefox the L3 headings are a lot smaller than the L1. Also Safari it looks good.

https://drive.google.com/file/d/1ulJ6HvIgcqGuugsQOXIhXBSBF7r2eRro/view?usp=sharing

organization/_posts/2018/2018-10-18-coordination.md
```diff
@@ -0,0 +1,155 @@
...
+
+# {{page.title}}
+
+*Present/Contributors*: Graeme Stewart, Andrea Valassi, Pere Mato,
+Daniel Elvira, Liz Sexton-Kennedy, Serhan Mete, David Lange, Ed Moyse,
+Heather Gray, Ian Collier, James Amundsen, Michel Jouvin, Tommaso
+Boccali
+
+News, general matters
+=====================
```
jouvin: You should not mix `====` heading styles with `#` as it is unclear which level is `===`. The end result is that every heading is a header level 1, making the page hard to read. I'd suggest to replace `====` by their `#` counterpart, with the header level reduced by one....

graeme-a-stewart: @jouvin - it really does **not** come out like that on any browser I try. The heading levels are well separated visually and `===` and `---` are [completely standard markdown elements](https://daringfireball.net/projects/markdown/syntax#header).

As I mentioned in my last comment, it's **pandoc** that selects this style of markdown, why would we want to change this by hand?

If you really insist I will change it, just to conclude on this, but it will not affect the rendered page in any way and more human changes mean more chances for mistakes...

jouvin: @graeme-a-stewart sorry I missed your previous comment, so my new one... sure the markdown syntax used by pandoc is perfectly valid, it was not my point. Even mixing it with the other syntax. Fixing by hand what is done by pandoc is non sense...

In fact the problem comes from the Google Docs where the headings are defined as level 1 where they should be level 2 IMO. It may be good to fix it once (I remember doing it in the past...) so that it is correct for next meetings.

The headings are well separated, just they are too big and this makes identifying the various part harders (at least for people like me!).

graeme-a-stewart: OK, sorry if I misunderstood then. (You know this reminds me of a gripe I have had forever, that in short documents the headers levels h1, h2, ... are all just way too large.)

Yes, I'd be fine with refining the 'source' headers in the live notes.

jouvin: @graeme-a-stewart I fixed the Google Docs (hopefully!). I let you redo the pandoc conversion and update the PR.

graeme-a-stewart: If we do this via the GDoc header level then all the new minutes are out of sync with the older ones. And actually I think there's maybe a better way. Isn't the right way to fix this to change the CSS for meeting minutes to make the font sizes more reasonable? That way all of the minutes are fixed post facto and we don't have to violate the correct *logical* structure of H1,2,3,...

I am no CSS wizard, so I might not be able to do this right away, but I think it's better than the other way.

jouvin: It is much easier in fact and documented at https://hepsoftwarefoundation.org/jekyll-beginners.html! We need to define the base header level in pandoc... Reverting GD changes...

jouvin: I also fixed in GD the pb with the bulleted h3...

eduardo-rodrigues:
Hi @jouvin, I did not follow this, hence the question: ready to be merged now?

PR #401, opened 2018-11-06
Fix of indentation in ordered list for TNs page

eduardo-rodrigues:
Apologies for the mistake. Should be OK now.


graeme-a-stewart:
Apologies - as reviewer I should also have spotted that. Thanks for the prompt fix.

PR #405, opened 2018-11-09
Weekly meeting minutes, 2018-11-08

graeme-a-stewart:


eduardo-rodrigues:
Just a detail: Travis fails as it does not find https://indico.cern.ch/event/762971/. Indeed the URL does not exist. Can you fix the link in the PR since you have a local checkout at hand?

graeme-a-stewart:
Thanks for spotting that - we cancelled that meeting and I was tidying up Indico to get rid of it.

The failure now seems to be some ephemeral issue with the BES3 website, so can be ignored.

PR #411, opened 2018-11-30
Website Updates

graeme-a-stewart:
A bit of a grab bag of updates:

- Weekly meeting minutes from 2018-11-29
- Modify front page centre column with new convenor appointments
- Announce our two meetings for next week (software forum, developer tools)

Separated into different commits, but I didn't think it was worth three separate PRs (can be squashed on accept).

### DIFF
--- a/_layouts/main.html
+++ b/_layouts/main.html
@@ -60,15 +60,15 @@ in High Energy Physics software and computing internationally.
  <div class="col-lg-4">
   <h2>New Working Groups</h2>
     <p>
-    The HSF is establishing new working groups with
-    <a href="/organization/working-group-mandates.html">mandates</a>
-    to pursue the goals from the
-    <a href="/organization/cwp.html">Community White Paper Roadmap</a>
-    published last year. We have setup a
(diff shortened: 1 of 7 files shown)


graeme-a-stewart:
Self accepting now as we need these updates

eduardo-rodrigues:
Apologies @graeme-a-stewart, I was just about to say that I would try and find time later today for the review.

eduardo-rodrigues:
I've just read the PR and didn't see much fishy apart from

IRIS-HEP would like a to have HSF -> IRIS-HEP would like to have HSF
Pete Elmer (PI) has proposed that Graeme take this on -> Pete Elmer (PI) has proposed that Graeme takes this on
Graeme will also give an [*Overview  -> Graeme will also give an [*WLCG Overview 
Licensing was be raised -> Licensing was raised

If you really want a suggestion ;-): how about we add for each new WG convener the experiment or group to which they are attached? I mean in the new WG pages. It's a nice way to show the breadth of the conveners selection.

Issue #413, opened 2018-12-05
rucio.cern.ch website check fails


graeme-a-stewart:
We get consistent failures for a few weeks now with the Rucio website:

```
- ./_site/gsoc/2018/proposal_ATLASrucio.html
  *  External link https://rucio.cern.ch/ failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: Peer certificate cannot be authenticated with given CA certificates
```

graeme-a-stewart:
Checking locally on my Mac I don't get the error. So it might be solved with a container update.

graeme-a-stewart:
So, I checked the container and noticed it contains a hand rolled certificate for `rucio.cern.ch`. If I remove that then the error reverts to `Peer certificate cannot be authenticated with given CA certificates`. Maybe that certificate is wrong, but my usual method of getting missing certificates now doesn't work. 

[https://whatsmychaincert.com/generate?host=rucio.cern.ch](https://whatsmychaincert.com/generate?host=rucio.cern.ch)

generates an error, `Protocol "ldap" not supported or disabled in libcurl`. *sigh*

The internet suggested to get the certificates with this SSL magic:

`openssl s_client -host rucio.cern.ch -port 443 -prexit -showcerts`

Finally, using the certificates from there (all the bits between `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`) seems to fix things.

The new container is pushed up to docker hub.

I document this *pain in the ass* partly for catharsis and partly so we can find it again in the future...

PR #415, opened 2018-12-05
Bundle of small fixes

graeme-a-stewart:
- Correct docker pull to pre-pull the correct container for the Travis
- Remove the unnecessary "data-proofer-ignore" for a few rucio.cern.ch links (fixed in #413)
- Update the CWP arXiv link for the Visualisation page
- Fix broken internal link from 2017 GSOC (seems to be found by the updated HTML proofer)


graeme-a-stewart:
Thanks - we got a stupid timeout again on the URL check, but the persistent problems are fixed.

PR #422, opened 2019-01-11
Small fixes for links that died

graeme-a-stewart:


graeme-a-stewart:
Self-accepted these small fixes.

PR #426, opened 2019-01-15
Added pointer to GSoC admin list rather than individual mails.

agheata:


agheata:
Fails due to link https://www.openml.org/ temporary not working, but this is not added by this PR. Tested locally and looks fine. Merging.

graeme-a-stewart:
Hi @agheata - the problem of false positives from the URL checker is a long standing one (we check > 500 external sites, so spurious failures happen). You did the right thing to ignore it. If a website fails more persistently then please create and issue and we will disable it.

PR #429, opened 2019-01-17
Added presentation on C&SW wrk recognition, citations

eduardo-rodrigues:
Following-up from our email exchange, @graeme-a-stewart.

I committed the PDF to the place where I could find other similar material given that the LHCb meeting is internal and I cannot link directly to it from the HSF website pages.

If you accept this before the creation of today's minutes the link can still go in the minutes :-).

Thanks.


organization/presentations.md
```diff
@@ -5,6 +5,10 @@ layout: default
 
 # HSF Presentations
 
+## 2019
+
+* [Computing & Software work - recognition & citations in publications](http://hepsoftwarefoundation.org/assets/EduardoRodrigues-LHCb-2019-01-15.pdf), LHCb Weekly Meeting, 15 January 2019, Eduardo Rodrigues
```
graeme-a-stewart: Could you make this an internal absolute link, rather than including the website name? (Which should be `https` in anywise.)

eduardo-rodrigues: Sure. I copied in fact looking at similar links. I may try and fix a few while at it …

One detail, though: what do you precisely mean by "absolute link"? Not `https://hepsoftwarefoundation.org/assets/EduardoRodrigues-LHCb-2019-01-15.pdf`, then? Just `assets/EduardoRodrigues-LHCb-2019-01-15.pdf`? Else I'm unsure.

graeme-a-stewart: Hi - yes, it maybe is wrong in quite a few places! It should be:

```
/assets/EduardoRodrigues-LHCb-2019-01-15.pdf
```

Internal = Implicitly refers to this website
Absolute = Starts with "/"

OK, I checked and this is still apparently still considered to be a "relative" link, which I find confusing, cf. a filesystem; I like my nomenclature better! But I suppose as I just invented it it was confusing ;-)

eduardo-rodrigues: OK 👍 ! I looked around and corrected the various occurrences. Hope all fine now. Thanks.

PR #431, opened 2019-01-18
Add arXiv links and some 2018 presentations

graeme-a-stewart:
Fill in the two latest CWP arXiv links
Add a couple of HSF presentations given last year

### DIFF
--- a/organization/presentations.md
+++ b/organization/presentations.md
@@ -11,6 +11,8 @@ layout: default
 
 ## 2018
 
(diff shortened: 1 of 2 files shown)


organization/presentations.md
```diff
@@ -11,6 +11,8 @@ layout: default
 
 ## 2018
 
+* [HEP Software Foundation](https://indico.ific.uv.es/event/3438/contributions/9763/attachments/6464/7304/COMCHA_HSF_Presentation.pdf), Spanish Computing Challenges Meeting (COMCHA), 12 September 2018, Graeme Stewart
```
eduardo-rodrigues: From memory it should be "CONCHA", which means "shell". Can you check?

eduardo-rodrigues: If that's the case then worth renaming the file too ;-).

eduardo-rodrigues: Just checked and you are correct :-). Sorry for the confusion.

PR #433, opened 2019-01-21
Update of PyHEP WG page

eduardo-rodrigues:


graeme-a-stewart:
Looks good to me. The only thing I don't like is that fact that `Python in HEP` appears inside quotes. I can't see why and it looks ugly IMO. As you are touching the file could you remove them?

graeme-a-stewart:
Thank you! Also looks better in the WG drop down now.

PR #449, opened 2019-01-30
Fix failing url by adding https

JavierCVilla:
`html-proofer` seems to fail if a link is not marked as `https`  or `http`


JavierCVilla:
Since this is minor fix and I got @graeme-a-stewart 's approval, I'm self-merging this PR.

PR #485, opened 2019-02-04
small rucio projects update

mlassnig:


agheata:
The list of mentors order seems to be wrong due to a previous commit, will fix later, this generates conflicts in other PR's anyway

PR #490, opened 2019-02-04
Fix the subscribe link to the generator WG

valassi:


valassi:
Self merging this trivial (and hopefully final) fix

graeme-a-stewart:
Yep, that got it! Thanks,

PR #491, opened 2019-02-04
Remove ucas lhcb 2016 job opening

valassi:


valassi:
Hi Graeme, I am simply removing a file that I needed up three years ago and ended up duplicated in the main directory and in the assets subdirectory.

By the way:
- do we have a policy for including job openings like this one on the web site? I am not sure assets is a good idea either...
- for this specific job opening, which was from 2016, I guess we could delete also the version in assets; I also cross checked that there is no link to it from inside the web site

This is not at all urgent.

Thanks Andrea

graeme-a-stewart:
> Hi Graeme, I am simply removing a file that I messed up three years ago and ended up duplicated in the main directory and in the assets subdirectory.

👍 

> 
> By the way:
> 
> * do we have a policy for including job openings like this one on the web site? I am not sure assets is a good idea either...

I would say we don't put job material on the website. People can send links to hsf-forum when they have something to advertise (Inspire HEP is a better platform for this, where job adverts are part of what they do).

> * for this specific job opening, which was from 2016, I guess we could delete also the version in assets; I also cross checked that there is no link to it from inside the web site

OK, then I would suggest that you do that also in this PR.

Annoyingly you still get these files in a clone but, hey ho, it's not that serious, so no need for the BFG...

> 
> This is not at all urgent.

Indeed!

valassi:
Hi Graeme, thanks a lot.
Ok so I removed also the job ad in assets.
And if I have any more job ads, I will only post them in the HSF forum.
Cheers
Andrea

PR #499, opened 2019-02-05
Add a script to convert docx to md

valassi:


valassi:
Hi Graeme, Michel,

I submitted this MR with a simple script I used for docx to md conversion and subsequent tidying up. I suggest you/we try it a few times first, and only merge this when/if ot seems appropriate and sufficiently solid. Let me know if you have other comments anyway!

By the way I eventually moved from python to sed/awk, and I also implemented one point that was on Graeme's todo list, for leaving empty lines before headers.

Thanks for the help!
Andrea

graeme-a-stewart:
Hi @valassi 

Super idea to improve this process with better docs and actually a script that people can use.

That said... I have to say I not really that keen on your version. 

- `mac2unix`, `dos2unix` - where to get these on a Mac or Windows machine?
  - Instead just use the `newline` argument to Python3's `open` function
- Converting to `sed`/`awk` is utterly retrograde IMO
  - The `sed` is more or less ok, the `awk` is just gobbledygook
  - Python is just way clearer for writing this stuff in a way that is compact, comprehensible, maintainable

I improved my own [pythonic converter](https://gitlab.cern.ch/graemes/utils/blob/master/hsf/pandocTidy.py) a bit, but it does make some mistakes when there are dashes in the markdown conversion, some of them get moved to the beginning of lines and become bullets. But it's 99% good...

Cheers

Graeme

jouvin:
@valassi I just used your script to convert the last coordination live notes and it worked well (where the plain pandoc producing a messy result). That said, I share Graeme's remarks and I let him decide what to do with this PR. We definitely need something asap...

I have another remark on your PR that makes it difficult to merge as it is: you merged the last master updated into your development branch... This is, IMO, a bad practice. This leads to a messy/hard to read history. The correct workflow should be to **rebase** your branch on the upstream/master. This way the master history remains linear and it is more inline with what you want to achieve: add your changes on top of upstream/master changes. Should this PR be merged, I suggest that you remove the commits after your last mod and then rebase your branch on upstream/master.

jouvin:
BTW, I had another look to the problem. I found in my test with the last live notes that the original problem seems to be solved if the GDOC is exported to epub instead of docx. The conversion seems fine in this case, without any need to tidying up... I've the feeling it is a better (and simpler) solution, avoiding potential future problems if the issue in pandoc conversion is  fixed.

graeme-a-stewart:
I think with the advent of web sites with direct editing of CodiMD the use of Google Docs + Pandoc isn't really what we should recommend any more. So I think this one we can close.

PR #501, opened 2019-02-05
Update HSF project text

graeme-a-stewart:
Emphasise more HSF hosting common software projects in it's role as a GSoC "project" entity

@JavierCVilla - do you think this improves things enough? I don't think it's necessary to add HSF as an organisation.


JavierCVilla:
> @JavierCVilla - do you think this improves things enough? I don't think it's necessary to add HSF as an organisation.

It looks great for me, better than before. Also, we do not have the problem with the organization anymore since `Phoenix` will appear as part of UMass and Pittsburgh. 

Thank you @graeme-a-stewart

PR #503, opened 2019-02-06
Proposal from ALICE.

agheata:
This is a proposal that I received by mail from Costin Grigoras from ALICE


agheata:
> I have again conflict with `mentors.md`, but otherwise it looks OK for me

We will fix these later on, for now let's keep git happy - the branch is mergeable. You maybe tried it from a branch, not from master.

PR #512, opened 2019-02-08
Changed link to article to point to open link version (L. Moneta)

agheata:


agheata:
I propose that if the change involves minor changes like this, like a link, a typo, ... to do the merge ourselves as soon as it passes the checks.

PR #521, opened 2019-02-20
 Remove "News" from the title (was "HEP Software Foundation News")

valassi:
It would be nicer if the tab title of our main page was "HEP Software Foundation" rather than "HEP Software Foundation News". I assume that index.html is where the change is needed (otherwise it is latest.html?).


graeme-a-stewart:
Hi @valassi - yes, well spotted. I completely concur with the update.

BTW, you seem to have many merge commits on this branch, but I am going squash them all down. In general merges should be avoided though (use rebase).

PR #550, opened 2019-03-27
Add minutes from packaging meeting of 2019-03-27

graeme-a-stewart:
@drbenmorgan - I had a spare 15 minutes, so I though it was a good idea to just get the minutes processed while still fresh in my head (and thanks for putting together a talk in a busy week for you)


drbenmorgan:
Thanks @graeme-a-stewart! Looks fine to me.

Issue #561, opened 2019-04-11
Link to nixcon2018.org does not work with https


JavierCVilla:
The continuous integration is consistenly failing due to this link:

`https://nixcon2018.org/` in this [post](https://github.com/HSF/hsf.github.io/blob/master/organization/_posts/2019/2019-02-27-packaging.md)

The url works without https though.

graeme-a-stewart:
Thanks - I'll fix it. It's very annoying when things like that happen, I had noticed it a while ago, but I usually leave it a little while to see if the website recovers.

JavierCVilla:
@graeme-a-stewart for some links it looks like just retrying the check works, I wonder the plugin offers such a feature.

graeme-a-stewart:
Believe me, if it offered something easy to use that made things more robust, we'd be using it already!

PR #562, opened 2019-04-11
Fix broken URL

graeme-a-stewart:
Drop https from nixcon (bad certificate)


graeme-a-stewart:
Should fix #561

graeme-a-stewart:
There is a problem with the chep2019 site right now, but I expect that will be fixed soon.

PR #572, opened 2019-05-13
Add Diego Davilla as co-mentor to CMS Tools proposal

sfiligoi:


JavierCVilla:
Looks good to me, sending an invitation link to Diego now to join to the organization in the GSoC Webpage.

PR #574, opened 2019-05-20
Add C++ Course

GuillermoFidalgo:


Schools/_posts/2019-05-20-Online-courseC++.md
```diff
@@ -0,0 +1,10 @@
+---
+title: C++ / Standard Template Library
+source: https://www-esh.fnal.gov/pls/cert/schedule.show_course_details?cid=11499
```
graeme-a-stewart: This link seems to be inaccessible unless you have a FNAL login - is there a public page instead?

GuillermoFidalgo: It does not. But looking at the page we've decided that this page does not have enough information for registration nor dates and there is no other official site. So it's not worth putting in the site.

PR #582, opened 2019-06-12
Add WLCG OB presentation by Michel

jouvin:


jouvin:
hepsoftware.org seems to have trouble again... Should we contact Torre?

graeme-a-stewart:
Thanks for contacting @wenaus about the hepsofware problems.

PR #593, opened 2019-07-13
Update front page and add next meeting

graeme-a-stewart:
Update the middle column of the front page to cover EPPSU (more recent
than HOW2019.

Add announcement for the next Reco & SW Trigger meeing.


_layouts/main.html
```diff
@@ -58,14 +58,21 @@ <h2>Meetings</h2>
...
     <p>
-      The HSF workshop at Jefferson Lab was a great success. Read our
-      <a href="/newsletter/2019/04/03/Jlab-Workshop.html">newsletter</a>
-      for an overview of all the topics we discussed.
+      HSF has been contributing to the 
+      <a href="https://europeanstrategyupdate.web.cern.ch">European Strategy Update</a> process,
+      emphasising the importance of software for the field.
+    </p>
+    <img src="/images/others/eppsu.png"
+    alt="EPPSU logo" style="display: block; width:40%; margin-left: auto; margin-right: auto;"></a>
```
jouvin: As said by the tests, there should be not '</a>' here...

jouvin:
@graeme-a-stewart there are 2 check errors. One is related to Codalab web site which is currently basically dead after a mistake done this week. We can probably live with the error for some time, they are working on restoring things... The other error is a real one, see my review.

_layouts/main.html
graeme-a-stewart: Yep - thank you link checker!

PR #595, opened 2019-07-16
Add Indico link for PyHEP 2019 workshop to WG page

eduardo-rodrigues:
This is the first bit of info on the workshop. A proper post will be added once the advertisement email goes out.


graeme-a-stewart:
All good @eduardo-rodrigues - we should do a front page announcement too.

PR #597, opened 2019-07-23
Made the list of schools into an include.

GuillermoFidalgo:
This PR is to add an include that uses the same code as before to make a list of the HSF training schools.
The major difference now is that it uses a workaround that I found to make a comparison between dates, something that jekyll seems to not like very much. 
This workaround permits the comparison of the "end date" of the school so that it can be properly classified into "Past School" or "Current or Upcoming School" instead of automatically changing the category of the school to Past School once the school has only started.

### DIFF
--- a/Schools/_posts/2019-06-06-art-of-data.md
+++ b/Schools/_posts/2019-06-06-art-of-data.md
@@ -1,8 +1,8 @@
 ---
 title: Gridka - The art of Data
 source: https://indico.scc.kit.edu/event/460/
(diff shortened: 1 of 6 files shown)


graeme-a-stewart:
Hi @GuillermoFidalgo - sorry for the late review, only just back from holidays!

GuillermoFidalgo:
I have applied the recommended fixes. Thanks for pointing out the mistake with the date of Gridka School. @graeme-a-stewart

PR #598, opened 2019-07-29
Twitter handle to PyHEP

tunnell:
Changed two files to add Twitter handle.


graeme-a-stewart:
Link would still be a good idea, but merged the original PR in anycase...

PR #603, opened 2019-09-13
Clarify status of page

graeme-a-stewart:
This is a historical page, so make that clear and provide a link to
the list of current working groups. At least one person reported
that a Google search sent them here, which was a confusing
place to end up...


graeme-a-stewart:
Thanks @eduardo-rodrigues

PR #605, opened 2019-09-26
Coord team members: remote P. Elmer's URL

jouvin:
Having a URL attached to a member is not a problem in itself but has the impact that the name is rendered in red, giving the impression they are the "bosses"! I think it is misleading and not inline with our way of working...

BTW, updating the coordination team page, it may be a good time to update the membership... Any suggestion?


graeme-a-stewart:
Yes, I agree - everyone can be found by Google these days, so no really need for person by person links anyway

PR #608, opened 2019-10-02
Deadline

GuillermoFidalgo:
Wanted to add some deadlines for the Upcoming Schools section of the Training Page. I managed to make it work but not without some weird Jekyll errors that don't seem to disrupt the generation of the page. 

**Would really appreciate some help with this**

### DIFF
--- a/Schools/_posts/2019-06-20-SOSC2019.md
+++ b/Schools/_posts/2019-06-20-SOSC2019.md
@@ -6,5 +6,6 @@ end_date: 2019-10-26
 author: Guillermo Fidalgo
 layout: event
 categories: Schools
+deadline: September 5
 ---
 [Third international School on Open Science Cloud](https://agenda.infn.it/event/19049/)
--- a/Schools/_posts/2019-10-21-INFNESC_School.md
+++ b/Schools/_posts/2019-10-21-INFNESC_School.md
@@ -6,5 +6,6 @@ end_date: 2019-10-26
 author: Guillermo Fidalgo
 layout: event
 categories: Schools
+deadline: September 16
 ---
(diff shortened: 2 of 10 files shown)


GuillermoFidalgo:
@graeme-a-stewart Could you take a look?

GuillermoFidalgo:
I did a little testing and it appears that the code itself is not the problem.
It looks like the problem is that this specific bit of code produces a parsing error when it runs **in the Training Page**. 
 If this code is run somewhere else like in the ` /Schools/events.html ` page it does not  produce an error.   😞

graeme-a-stewart:
Hi @GuillermoFidalgo 

So what I get is:

```
             Error: A Liquid tag in the excerpt of _workinggroups/training.md couldn't be parsed.
             Error: could not read file /Users/graemestewart/docs/hsf.github.io/_workinggroups/training.md: undefined method `ancestors' for nil:NilClass
```

Is that what you see? I'm very reluctant to merge code that has errors in it...

Sorry it took so long to have look!

Graeme (Happy New Year!)

GuillermoFidalgo:
Yes this is what I see. Whats wierd to me is that the error message does
not appear if that same code is located on a diferente page. The
/school/events.html for example does not show this message if I copy and
paste the code here.

graeme-a-stewart:
Hi @GuillermoFidalgo - I really lack the time to dig into this right now. Maybe @hegner could take a look and see if he has any insight.

graeme-a-stewart:
Actually, could you factorise the updates for schools themselves (which we should try to take ASAP) from the infrastructure changes where we're having trouble? Thanks.

PR #616, opened 2019-11-21
Update to minutes page

eduardo-rodrigues:
@graeme-a-stewart, this is just a trivial update in wording, startup -> coordination, which I noticed while looking for the last minutes … The minutes number 174 are missing; I believe they correspond to the minutes taken with CodiMD ...

### DIFF
--- a/organization/minutes.md
+++ b/organization/minutes.md
@@ -7,7 +7,9 @@ layout: default
 
 ## Meeting Minutes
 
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Hi @eduardo-rodrigues about the minutes, I think there are none missing - it was me miscounting when I changed the title on the Google Doc (it had been bumped up by 1, even though we didn't use it for no. 173, but used CodiMD instead).

I'd still like to take the minutes in CodiMD, but we need feedback if CERN would allow lightweight/EduGAIN accounts to contribute (or we use the CodiMD demo service).

eduardo-rodrigues:
The interface in CodiMD is indeed nicer.

PR #618, opened 2019-11-28
Announce next simulation WG meeting

graeme-a-stewart:


graeme-a-stewart:
Some URL failures that need to be fixed, but not related to this commit - self merging

PR #625, opened 2020-01-06
Rename logos according 2020 rules

aflp91:
Rename previous logos according to new 2020 name rules as written in new instructions.


aflp91:
Well … in fact I'm realizing there is more stuff to do: I should update previous texts referring to the old logos … Not sure is worth it. 

Maybe keeping the new png logos despite apparent duplicates?

aflp91:
I've reverted to keep previous logos (2019 pages seem OK) and added new right formatted logos from all previous ones.

PR #650, opened 2020-01-21
Add training schools with working logic

GuillermoFidalgo:


klieret:
``training.md`` should be the same as the life version now, right (that means only the new events have been added)? (unfortunately github handles these diffs a bit poorly, see https://stackoverflow.com/questions/16306012/)

graeme-a-stewart:
Yes, I think it's a GitHub booboo that it shows a differnce here, when in fact the file is the same as the master branch, which I checked:

```
(base) sapphire:hsf.github.io$ git diff gf/add-training-schools master -- _workinggroups/training.md
(base) sapphire:hsf.github.io$ 
```

So looks good...

PR #657, opened 2020-01-27
Change title of training school page

klieret:
We just discussed this in the training meeting: https://hepsoftwarefoundation.org/Schools/events.html is titled "Schools for HSF Training", even though only a subset of the schools advertised has us involved at all. Rather, this is meant as a list of any event concerened with HEP Software training.

Hence a title change.


graeme-a-stewart:
Thanks @klieret!

PR #671, opened 2020-01-30
Move Schools from markdown to yml file

GuillermoFidalgo:
https://github.com/HSF/hsf.github.io/issues/651

Issue #651, opened 2020-01-21
Simplify School template


klieret:
Currently the template for adding new training Schools looks like this:

```markdown
---
title:  Alpaka Parallel Programming - Taster Session and Basic Tutorial
source: https://indico.cern.ch/event/858758/
date: 2020-04-27
end_date: 2020-04-27
author: Kilian Lieret
layout: event
categories: Schools
---
[Alpaka Parallel Programming - Taster Session and Basic Tutorial](https://indico.cern.ch/event/858758/)
```

so the body text is always ``[title](source)``.

I think we should automatize this (and not have any body text at all), i.e. only have

```markdown
---
title:  Alpaka Parallel Programming - Taster Session and Basic Tutorial
source: https://indico.cern.ch/event/858758/
date: 2020-04-27
end_date: 2020-04-27
author: Kilian Lieret
layout: event
categories: Schools
---
```
because

1. this avoids copy-paste mistakes
2. this is more future proof, e.g. if we want to change the way the links are displayed consistently (add deadlines etc.)

@GuillermoFidalgo

klieret:
Actually looking at ``list_of_schools.md``, it looks like the body of the template isn't even used anyway, right?  
Or is that used somewhere else?

```markdown
{% assign schools = site.categories.Schools | sort:"date" %}

{% capture now %}{{'now' | date: '%s' | plus: 0 }}{% endcapture %}

## Current and Upcoming Training Schools
#### **Warning** : Application deadlines are **before the date shown**
{% for post in schools %}
  {% capture date %}{{post.end_date | date: '%s' | plus: 0 }}{% endcapture %}
  {% if date > now %}
  1. [**{{post.date | date: "%-d %b"}} - {{post.end_date | date: "%-d %b %Y"}}** - {{post.title}} ]({{post.source}})
  {% endif %}
{% endfor %}

## Past Schools
{% for post in schools reversed %}
  {% capture date %}{{post.end_date | date: '%s' | plus: 0 %}}{% endcapture %}
  {% if date < now %}
  1. [**{{post.date | date: "%-d %b"}} - {{post.end_date | date: "%-d %b %Y"}}** - {{post.title}} ]({{post.source}})
  {% endif %}
{% endfor %}
```

if we just replace

GuillermoFidalgo:
Hi @klieret thanks for opening up this issue.

I agree that this `[text] (url)` in the body of the md files be automated.

[...]

PR #692, opened 2020-02-04
Add activity script

klieret:
Recently (#671) we changed the way we add training schools to the list at https://hepsoftwarefoundation.org/Schools/events.html as discussed in #651 .

This adds a small python script with a command line interface to more conveniently add a new event to the yml database and checks that it makes sense and brings the yml file in a uniform order.


_data/training-schools.yml
```diff
@@ -1,179 +1,177 @@
-- title: AEPSHEP (Asia-Europe-Pacific School of High-Energy Physics)
```
klieret: Nothing happens in this file, it's just sorted by date and the dictionary keys are sorted alphabetically; all blanks are removed.

klieret:
@GuillermoFidalgo  Can you try if this also works for you: Just execute ``scripts/add_event.py`` and it should guide you through adding a new event

GuillermoFidalgo:
Sure

klieret:
As I have tested the script several times and it doesn't change the webpage itself, I will proceed and merge this; we can still modify/roll back later.

Because it reorders the yml file and I want to add new events (which risks merge conflicts), I don't want to leave this PR pending.

Hope that's alright for everyone.

PR #718, opened 2020-02-10
Fix broken link

aflp91:
Starting fix for GSoC broken links

### DIFF
--- a/announcements/_posts/2020-01-07-gsoc.md
+++ b/announcements/_posts/2020-01-07-gsoc.md
@@ -3,7 +3,7 @@ title: HSF GSoC 2020 call for projects and mentors
 author: Andrei Gheata
 layout: newsletter
 symbol: glyphicon-calendar
-link: activities/gsoc.html#for-projects-and-mentors
+link: activities/gsoc.html
 until: 2020-02-05
 ---
-[HSF GSoC 2020 call](activities/gsoc.html#for-projects-and-mentors)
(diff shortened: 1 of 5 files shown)


announcements/_posts/2020-01-07-gsoc.md
```diff
@@ -3,7 +3,7 @@ title: HSF GSoC 2020 call for projects and mentors
 author: Andrei Gheata
 layout: newsletter
 symbol: glyphicon-calendar
-link: activities/gsoc.html#for-projects-and-mentors
```
agheata: This was a valid link (was actually working), it is the checker that is wrong...

agheata: But we can change it indeed to reduce the noise from the checker.

jouvin:
The checker always had a few problems but it is not a good practice to ignore the errors. Looking at the errors, I think that most of them are real ones. Those related to https links not responding, in particular the Rucio one, reminds me a problem but I forgot the details. May be @graeme-a-stewart remembers... 

I saw one error related to a mailto URL which is not correct IMO. I have not checked the source file but I think the `mailto:` prefix may be missing.

graeme-a-stewart:
Re. the plugin errors, there are indeed many real errors that need to be checked and cleaned up. I started to do it, but it's rather tedious and slow.

The underlying problem is that we have 1448 links that get checked so there are always spurious errors from sites that timeout or are temporarily down, so we really need to rethink this infrasctucture.

PR #740, opened 2020-03-08
Fix proposal typo.

vgvassilev:


agheata:
This does not show the diff compared to the previous version (it is like this is a new file and it isn't), so I fail to see the modification.

vgvassilev:
Sorry, forgot to push the full set of changes... Should be fixed now.

PR #741, opened 2020-03-09
add MCnet summer school

GuillermoFidalgo:
Added one more school. 

Also this is the first time using a script by @klieret. Works great!


klieret:
Thanks :)

PR #744, opened 2020-03-10
Add ical link and use inline version of the calendar by default

graeme-a-stewart:
ical link is needed for people to import the calendar feed

inline version keeps the "theme" of the website, but we give a link to
the single page google version of the calendar


graeme-a-stewart:
Some small improvements in how we present the community calendar information

graeme-a-stewart:
Good point - let me see if I can address that. In *ye olde days* the idea was that the Google Calendar would get all the events from Indico, but we didn't get there and it's redious and error prone doing this by hand - @roiser was going to take a look, #725.

PR #736, opened 2020-02-28
Fix all persistently broken URLs

graeme-a-stewart:
Trawled through all broken URLs on the site and redirected, disabled or just removed them, as necessary.

There is still an issue with the `rucio.cern.ch` certificate, that required manual patching in the docker image (<https://github.com/HSF/docker>). New image was pushed to docker hub, so should be running fine now.

Fixes #613 

Just to prove it worked:

```
Running ["ImageCheck", "HtmlCheck", "LinkCheck", "ScriptCheck"] on ["./_site"] on *.html... 

Checking 1537 external links...
Ran on 792 files!

HTML-Proofer finished successfully.
```

PR #745, opened 2020-03-10
Add note on training events with un-validatable URLs

graeme-a-stewart:
@klieret - this issue came up with the Trans-Siberian School when I was fixing all of the persistent bad links in #736.

Note the supporting changes that were made:

- <https://github.com/HSF/hsf.github.io/pull/736/files#diff-86ad63be0256d177ec9de64b737d3e8b>
- <https://github.com/HSF/hsf.github.io/pull/736/files#diff-7f814d9e5bdc111dc5c9cb827af272a7>


klieret:
Hi! Thanks for documenting this. I've also already updated my ``scripts/add_training_school.py`` script to not complain about the extra property :)

Issue #752, opened 2020-04-09
How to include additional pages for workinggroups


klieret:
We want to add additional pages for the training working group. What's the best approach? I could 

# Option 1
change it so that we can have subpages in the workinggroups directory, i.e. 
```sh
cd _workinggroups
mkdir training
mv  training.md training/index.md
touch training/something.md
```
and change the permalink of the workinggroups from ``/workinggroups/:title.html `` to ``/workinggroups/:path.html``  

# Option 2
Have everything training related go to ``hepsoftwarefoundation/trainingdocs/..``, i.e. 
```sh
mkdir trainingdocs
```
and add it to ``_config.yml``. 

# Option 3
Have everything training related (including the training home)  in ``_training`` and set it up so that ``hepsoftwarefoundation.org/training`` goes to the index and ``hepsoftwarefoundation.org/training/something`` goes to the respective pages.
I'd add an additional redirect from ``hepsoftwarefoundation.org/workinggroups/training`` to ``hepsoftwarefoundation.org/training`` to not break any links.

klieret:
I personally favor the option 3 because of the short and concise URL.

graeme-a-stewart:
Hi @klieret - I think Option 3 sounds generally good. However, I think that a landing page for the Training WG needs to be maintained in `_workinggroups/training.md` as the set of pages here are use dynamically in a few places:

https://github.com/HSF/hsf.github.io/blob/3acabc613a47db88b22093d948602e12d494c37c/_includes/navbar.ext#L29

So if it was being redirected it would get lost.

klieret:
Thanks! Let's review the details on #757

PR #757, opened 2020-04-17
Training add more pages

klieret:


_training/curriculum.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Towards a HEP Software Training curriculum"
+layout: plain
+---
+
+<style type="text/css">
```
klieret, line 6: Probably not the best style to just paste this here, but seems to work well.

klieret:
Hi @graeme-a-stewart . Thanks your feedback on #752. I tested this setup locally and it seems to work well. What do you think?

_training/curriculum.md
graeme-a-stewart: I don't think it's so important!

graeme-a-stewart:
Yes, I think it looks good - content and technical solution.

Shall I merge it now?

klieret:
Thanks @graeme-a-stewart

PR #771, opened 2020-06-08
Improve website link reliability

graeme-a-stewart:
Remove the hepsoftware.org links, has been down for quite a few weeks. We have to admit this just didn't get traction..

To be worked on: better analysis of URL failures to only really FAIL when connected to a particular PR

### DIFF
--- a/.travis-scripts/html-proofer
+++ b/.travis-scripts/html-proofer
@@ -17,6 +17,9 @@ then
 fi
 
 # DESY Debug statetments
+echo NETWORK
+ifconfig -a
+echo
 echo HTTP HEADER TEST
 curl -I -L https://indico.desy.de/indico/
 echo 
--- a/.travis-scripts/html-proofer
+++ b/.travis-scripts/html-proofer
@@ -16,6 +16,14 @@ then
     usage
 fi
 
+# DESY Debug statetments
+echo HTTP HEADER TEST
+curl -I -L https://indico.desy.de/indico/
+echo 
+echo HTTP PAGE TEST
+curl -L https://indico.desy.de/indico/event/22731/
+# End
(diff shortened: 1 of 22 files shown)


graeme-a-stewart:
Note there seems to be an issue with DESY's Indico in particular - I asked some DESY colleagues about this, so waiting for feedback on that

graeme-a-stewart:
OK, so I am really not sure why the CI isn't running for the later commits (issue with applying the WIP label?)

graeme-a-stewart:
> Note there seems to be an issue with DESY's Indico in particular - I asked some DESY colleagues about this, so waiting for feedback on that

Ticket was sent to DESY's helpdesk with the current status
- From k-net ISP (my home) indico.desy.de gives 403 (Forbidden) errors
- From inside a Travis CI job the connections just timeout (tested directly with curl)
- From CERN everything works ok (although the HTTP response code is 400, BAD REQUEST)

Looking back through the CI logs, it seems the DESY issues started about a month ago.

graeme-a-stewart:
Just to note that updating the hsf-jekyll container right now results in some serious error rates (all the SSL websites seem to fail). This probably needs a serious re-examination, with <https://github.com/gjtorikian/html-proofer/wiki/Using-HTMLProofer-From-Ruby-and-Travis> in mind.

graeme-a-stewart:
DESY update:

- k-net block was removed
- still looks like packets from Travis don't arrive and DESY say they don't block them

graeme-a-stewart:
I am going to accept this one now as it at least cuts the errors down to just 3.

These are all still related to DESY Indico, which is giving 403s when accessed from Travis. To be debugged later...

PR #773, opened 2020-06-24
Create robin_newhouse.md

robinnewhouse:


klieret:
Thanks a lot @robinnewhouse . For the gravatar, I'll replace your email with the hash later (but it won't matter because the github profile picture is taken over it anyway).

PR #781, opened 2020-06-25
Bug in homepage link

henryiii:
The homepage link should take you to the homepage given, not just back to the person page again.


klieret:
Good catch! Thank you :)

PR #782, opened 2020-06-26
Add edit on github button

klieret:
This makes it easy for people to see the markdown source and send around change proposals, even if they don't know how to do git. People who have write permission on the repo can also directly do changes through the github web interface in a seconds notice.

image

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -22,7 +22,8 @@
 <br><br>
 <hr>
 <div class="footer fixed-bottom">
-Thanks to <a href="https://pages.github.com/">GitHub Pages</a>, <a href="http://jekyllrb.com/">Jekyll</a> and <a href="http://getbootstrap.com/">Bootstrap</a>
+<a href="https://github.com/HSF/hsf.github.io/edit/master/{{ page.path }}"><i class="glyphicon glyphicon-wrench"></i> Improve this page. </a>
+Thanks to <a href="https://pages.github.com/">GitHub Pages</a>, <a href="http://jekyllrb.com/">Jekyll</a> and <a href="http://getbootstrap.com/">Bootstrap</a>.
 </div>
 
 </div> <!-- container -->


graeme-a-stewart:
Hi @klieret 

This is a nice idea thanks! Do you know how this looks for someone who does not have developer rights for the repo? I get the option to directly edit the master branch (not even going via a feature branch), which is quite dangerous. But I am an admin, so maybe that's just for me...

We got into trouble in the past with people screwing up formatting during direct commits to master.

Graeme

image

klieret:
Ah, sorry, merged before you finished your comment. Let me check

klieret:
I also can directly commit to the master branch (but I also have direct push permission, so I think that makes sense). Let me try how this looks for an outsider

graeme-a-stewart:
> 
> 
> Ah, sorry, merged before you finished your comment. Let me check

I think it's ok - changing this will be somewhere in the project settings... I did add the constraint that 1 review is required for merges to master. But, `sudo make me a sandwich` is always possible

klieret:
That means that normal pushing to master from the CLI is also not possible anymore, right? While I understand this action, it still makes trivial content updates a bit more cumbersome of course...

graeme-a-stewart:
> 
> 
> That means that normal pushing to master from the CLI is also not possible anymore, right? While I understand this action, it still makes trivial content updates a bit more cumbersome of course...

If you are an admin then you can still overrule and merge without approval, but we do want to limit that to a trustable few...

graeme-a-stewart:
OK, my son helped me check what the edit looks like for people outside the project and it's exactly what we want:

WhatsApp Image 2020-06-26 at 12 16 09
WhatsApp Image 2020-06-26 at 12 16 09 (1)

klieret:
Looking back, I think this discussion was really missing some programming memes (but I'm happy with the speed in which PRs are reviewed recently).

image

PR #786, opened 2020-06-29
Training content

klieret:


klieret:
Those are some very boring changes, @graeme-a-stewart , could you take a look?

PR #787, opened 2020-06-29
Add Clemens Lange as educator

clelange:
Adding myself. Thanks!

### DIFF
--- a/_educators/clemens_lange.md
+++ b/_educators/clemens_lange.md
@@ -1,0 +1,19 @@
+---
+# Required:
+title: "Clemens Lange"
+country: FR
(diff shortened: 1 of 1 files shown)


clelange:
@klieret - I think it would make more sense if the gravatar picture was used by default (if provided), and only then the github picture, since the only purpose of the gravatar is the picture. The order of preference is correctly stated in the documentation, but I would suggest to change it.

klieret:
Thanks for the suggestion @clelange, this does make a lot of sense. I will include it in the changes when I'll generalize the profiles to be more useful to all of HSF :)

PR #791, opened 2020-07-03
add kevin

klieret:
@kmnelson

### DIFF
--- a/_educators/leonora_vesterbacka.md
+++ b/_educators/leonora_vesterbacka.md
@@ -1,0 +1,17 @@
+---
+title: Leonora Vesterbacka
+country: SE
+roles: [facilitator,mentor]
+years: [2020]
+
+github: leonoravesterbacka
+gitlab: mvesterb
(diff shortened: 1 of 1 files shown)


_educators/kevin_nelson.md
```diff
@@ -0,0 +1,22 @@
+---
+# Required:
+title: "Kevin Nelson"     # first (middle) last
+country: US              # your country of residence (2 letters)
+roles: [Instructor]              # subset of [facilitator, instructor, mentor], can stay empty ([])
```
klieret, line 5: Currently this is still case sensitive, but I can also fix that later for you after merging this PR :)

graeme-a-stewart: So shall I just merge this now, right?

klieret: Yes, actually it already works as is, will just fix it later for consistency

klieret:
Thanks!

PR #806, opened 2020-07-17
Update procedure for HSF letters

graeme-a-stewart:
As agreed in the coordination meeting of 2020-07-16:

* WG convenors should also be involved in the discussion
* Contact with coordination is often more informal so be a little less specific about an "email"


graeme-a-stewart:
I will just merge this one as it's a rather trivial change and we agreed to it in the coordination meeting of 2020-07-16

PR #809, opened 2020-07-23
Training sitemap

klieret:
A sitemap rather than lots of links hidden in the text.

looks like this:

image


hegner:
Could you fix
    *  internally linking to /training/educators, which does not exist (line 172)
359     <a href="/training/educators">This page</a>
360  *  internally linking to howto-new-module, which does not exist (line 170)
361     <a href="howto-new-module">here</a>
362  *  internally linking to howto-update-module-style, which does not exist (line 170)
363     <a href="howto-update-module-style">here</a>
364- ./_site/training/educators.html

?

klieret:
Ah right, I forgot about this quirk from the link checker. The pages are accessible like this, e.g. [educators](hepsoftwarefoundation.org/training/educators), but the link checker wants the ``.html`` suffix (which makes sense I guess)

hegner:
yes, but we have so many false positives in external links that we should keep internal checks clean

klieret:
Ok, I'll fix this later, thanks for pointing this out!

PR #821, opened 2020-08-20
Pretty stati

klieret:
image

As a side note: I think the source will actually become more readable in HTML at this point, but I'd open a separate PR for that.


graeme-a-stewart (approved):
Nice!

PR #825, opened 2020-08-27
Add some 2020 HSF talks

graeme-a-stewart:
Prompted by #823 I realised I should add a few other talks from 2020


eduardo-rodrigues (changes requested):
I was indeed surprised to see me adding the first 2020 presentation ;-).

Graeme, are your presentations in public agendas? If that's the case you could just have the direct link since one can then easily go to the agenda by removing some bits of the URL? I actually wonder if we should now start adding links to the agendas, where possible ...

In any case, if you add the PDFs, I would add your name to the filename, as done in general, for bookkeeping.

eduardo-rodrigues:
BTW, if you think it is a good idea to add agenda links when possible, would you be so kind and add mine for the EIC talk (https://indico.bnl.gov/event/9122/)? Thanks a lot.

graeme-a-stewart:
Hi @eduardo-rodrigues - I updated to add direct links to the Indico agendas instead of local PDFs (I can never really decide what's best!). I also added the seminar page link for your talk.

eduardo-rodrigues (approved):

PR #843, opened 2020-10-12
Profiles generalize #835

### LINKED ISSUE #835: Floating heads for WGs
We agreed at the HSF coordination meeting that it would be good to implement the floating heads for the convenors of all the WGs, generalising what was done for the training group.

@klieret as the person who did the nice training implementaion can you see how to implement this easily?

klieret:
This took longer than I thought. But now it should not be training specific anymore (should have really thought of this from the beginning).

Closes #835


klieret:
**I think this should not be squashed**, I already tried to squash some things locally, but putting this all in commit will be a complete mess, because too many things were changed, many automatically.

graeme-a-stewart:
Hi @klieret 

So, I am having a very odd error when trying to check this PR for you, I'm getting:

```
bash-5.0$ bundle exec jekyll build
Configuration file: /srv/jekyll/_config.yml
            Source: /srv/jekyll
       Destination: /srv/jekyll/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
jekyll 3.8.5 | Error:  No such file or directory @ rb_sysopen - /srv/jekyll/_site/schools/2020/06/07/tCSC.html
```

At the moment I'm working out of a container as Ruby seems to have got into one of its traditional funks on my laptop.

`docker run -t -i -v $(pwd):/srv/jekyll --user $(id -u) --env HOME=/tmp hepsoftwarefoundation/hsf-jekyll bash`

Are you seeing anything like this?

klieret:
Hmm, I cannot reproduce this. Here's what I did to ensure I was doing the same thing as you:

```bash
docker pull hepsoftwarefoundation/hsf-jekyll
cd /path/to/my/cloned/repo
docker run -t -i -v $(pwd):/srv/jekyll --user $(id -u) --env HOME=/tmp hepsoftwarefoundation/hsf-jekyll bash
bundle install
bundle exec jekyll build
```

klieret:
Hi @graeme-a-stewart @hegner . Ideally we shouldn't keep this dangling for too long, because people will soon open PRs for their contributions via the C++ or the GPU/ML school (which would then have to be reworked again)

klieret:
Okay, so we're too late ;)    In that case **please don't merge this directly after approving**, because I will need to merge master and update the new profiles before this goes life!

graeme-a-stewart:
Hi @klieret 

Yeah, sorry this then just slipped my mind when I could not check locally. Ruby/bundler is still broken for unfathomable reasons on my Mac, but I used another machine.

So I get weird things now. At the bottom of the training WG page there's a trailing `</div>` and for the training community I only get a `</div>`.

So something isn't right yet...

image
image

graeme-a-stewart (changes requested):
See comment for how this appears broken for me - not quite clear where it's going wrong

klieret:
I have trouble reproducing this. I used the docker container to build and then from outside the container ran `jekyll serve --skip-initial-build`  ([can't directly access localhost from within the container](https://stackoverflow.com/questions/20813486/exploring-docker-containers-file-system)), but everything looks fine there.

klieret:
Let me try on another machine

klieret:
I can't reproduce this on the other machine either.

The trailing `</div>` is a bug that can happens when there are no profiles found at all. But why would that be the case for you?

graeme-a-stewart:
So, very strange - I switched branches from master and jekyll produced that result. However, if I start jekyll already on this branch it all works properly.

Ah, I understand why - you change `_config.yml` and jekyll doesn't re-read that... OK, yes, now I see that it's good.

klieret:
Can you check whether you can access `http://127.0.0.1:4000/profiles/kilian_lieret.html` or `http://127.0.0.1:4000/training/profiles/kilian_lieret.html` or check the output of `http://127.0.0.1:4000/profiles/` or `http://127.0.0.1:4000/training/profiles/` (I just removed the `training` bit of the link in the last commit)

graeme-a-stewart (approved):

klieret:
Perfect, then I'll update the two profiles that have been added since and will merge :)

klieret:
(Had to squash in the end, because rebasing apparently doesn't work)

klieret:
Looks like everything is working, but let's leave the branch for a little bit longer

PR #854, opened 2020-11-17
add iDDS

wguanicedew:


_activities/idds.md
```diff
@@ -0,0 +1,25 @@
+---
+title: intelligent Data Delivery Service
+author: Wen Guan
+layout: plain
+---
+
+## Introduction
+
+iDDS(intelligent Data Delivery Service) is an intelligent Data Delivery Service. It's an ATLAS/IRIS-HEP joint project. It’s designed to intelligently transform and deliver the needed data to the processing workflow in a fine-grained approach for High Energy Physics workloads. It will not only reduce the need for replicas, but also enable benefits such as decreasing the time period of caching transient data, transforming expensive replicas to cheaper format data at remote sites and only cache cheaper new data for processing. iDDS will also work to orchestrate the WorkFlow Management System (WFMS) and the Distributed Data Management (DDM) systems to trigger them to process the data as soon as possible. In addition, iDDS will have intelligent algorithms to adjust the lifetime of cache, the format transformation and the delivery destination. The iDDS will increase the efficiency of data usage, reduce storage usage for processing and speed up the processing workflow.
```
graeme-a-stewart, line 9: The first sentence here seems to have an extra *intelligent Data Delivery Service* in it. I think would be better as

> iDDS is an *intelligent Data Delivery Service*

I would also avoid "it's" - the page should be a little more formal... so "it is" please!

The 3rd sentence is quite long and I think is clearer if you punctuate it like this (and add a few articles, as below):

> It will not only reduce the need for replicas, but also enable benefits such as: decreasing the time period of caching transient data; transforming expensive replicas to *a* cheaper format data at remote sites; and only cache *the* cheaper new data for processing

I think it will be easier on the eye if you put a paragraph break after that sentence, at the start of "iDDS will also work to orchestrate the WorkFlow...".

And "lifetime of cache" should be "lifetime of *the* cache".

_activities/idds.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Main description
+
+The iDDS is designed as a standalone experiment agnostic service. It's implemented based on a Work/Workflow structure, where a Work is a transformation/processings and a Workflow is a bunch of Works and their relationship, such as DAG. New use cases can be imported to iDDS by implementing a new Work with some hook functions.
```
graeme-a-stewart, line 13: It's -> It is

a Work -> Work

bunch of -> collection of ("bunch" is rather informal)

"such as a DAG" - such as or always as a DAG? I'm trying to imagine what wouldn't be a DAG.

a new Work -> a new Work Item (it's a bit awkward as "work" is normally an uncountable noun in English, so to use it as the name for countable items grates a bit).

graeme-a-stewart (changes requested):
Thanks a lot Wen. I made a few suggestions to improve the way the text reads, but it's a very good project description - just what we were missing.

_activities/idds.md
wguanicedew: I updated to fix them.

graeme-a-stewart (approved):
Thanks @wguanicedew (sorry for re-reviewing late - workshop took all my attention!)

PR #860, opened 2020-11-27
Swap WGs and Activities

graeme-a-stewart:
Working groups are rather more important
in the HSF, so swap the order on the web page
so that they come first


klieret (approved):
looks good to me!

PR #862, opened 2020-12-02
Centre bar advert for training hackathon

graeme-a-stewart:
Good to advertise this as our end of year party!


klieret (approved):
Coooool!

image

PR #865, opened 2020-12-21
Increase visibility of link to coordination team

klieret:
I think it's a bit hard to spot the link to the coordination team page. It should be a bit more prominent than it currently is IMO.

But not feeling too strongly about this either, just took me a while to find it


graeme-a-stewart (approved):

graeme-a-stewart:
Hi @klieret - it seems like a good idea to have this a bit more prominent, so happy to accept the change.

Thanks!

Graeme

PR #875, opened 2021-01-20
Add documents page

graeme-a-stewart:
Add a missing page - all of the documents that we have written as the HSF (as noticed by @peremato).

Reorganise the "Communication" drop down a bit to emphsise the more active/useful links.


eduardo-rodrigues (approved):
Very good idea and update :-)!

graeme-a-stewart:
Thanks @eduardo-rodrigues - I'll pop this in before the meeting so I can also publicise it there

PR #880, opened 2021-01-25
Add new link checker test

### LINKED ISSUE #184: URL checking is unreliable
As discussed in https://github.com/HEP-SF/hep-sf.github.io/pull/183 with @hegner and @jouvin , we now have so many external URLs in the HSF website we observe a significant failure rate from

- Websites that are really in downtime for some ~hours
- Temporary timeouts

Immediate retries would help with the later. A nightly check might help with the former.

However, we should be aware that over complicating the infrastructure for doing this would not be a good solution.

### LINKED ISSUE #869: Migrate website proofer to github actions
Travis is going almost fully commercial and will only offer limited support to open source projects in the future.

I have therefore disabled the Travis webhooks for hsf.github.io, which has disabled Travis runs. The reality is that the link checking became so unreliable (see #184 ) that it was practically useless. In addition, running Travis via webhook was deprecated anyway, in favour of GitHub Apps.

Moving to GitHub actions allows us to look again at how the URL checker runs and to try and improve it to be more reliable.

graeme-a-stewart:
Add a new GitHub Action based link checker
that will validate only the changed files in
each PR (so is faster and less error prone than
previous checks of everything)

Closes #184 
Closes #869


graeme-a-stewart:
This PR implements a new link checker for the HSF website. It seems to do pretty much what we want - only updates are checked, so we avoid the false positives that we had with the old checker (which checked everything, and that was more than 1000 links).

In parallel we should probably run a weekly check aginst all links on the site to clean up things that have died since they were referenced.

graeme-a-stewart:
Updated YAML to also check on `pull_request` to get the check run as part of the PR itself.

graeme-a-stewart:
P.S. Just to prove that this would actually work, I deliberately inseted a dead link and the check did then fail:

https://github.com/graeme-a-stewart/hsf.github.io/actions/runs/510160953

klieret (approved):

klieret:
Nice! Did you also check what happens if you run it accros all files? There were always particular cases that the last link checker had its problems with (linkedin links, etc.).

klieret:
In either case, the fact that it only takes the changed files is a real game changer if we can't completely exclude the possibility of false-postives

graeme-a-stewart:
> Nice! Did you also check what happens if you run it accros all files? There were always particular cases that the last link checker had its problems with (linkedin links, etc.).

I did -  there are some things it has picked up that should be fixed. However, it is quite a lot less efficient than the old link checker as it checks each MD in turn, separately, so if a URL is referenced `N` times from different places then it gets checked `N` times.

The old checker definately had some issues that were to do with IP blocking from the Travis build nodes, so it may be that this is improved just by moving to GitHub actions.

graeme-a-stewart:
> In either case, the fact that it only takes the changed files is a real game changer if we can't completely exclude the possibility of false-postives

Having looked at that over the years that I have been maintaining the website it's clear that a single check of 1000 different URLs is almost inevitably going to have a couple of spurious failures. One of the better things with this link checker is that retries are built in and configurable.

hegner (approved):

hegner:
To me that looks very great. If Graeme confirms, I will merge it.

graeme-a-stewart:
Thank you both @klieret and @hegner. Yes, I think this should be merged now and then we keep an eye on it for teething issues.

@hegner if you merge I would also squash as the intermediate commits are not interesting.

PR #886, opened 2021-02-02
Update minutes with Frameworks WG plans

graeme-a-stewart:
So that the minutes become a record of the plans for 2021, add some ideas from the frameworks group


graeme-a-stewart:
Trivial update, so self-approving

PR #890, opened 2021-02-04
VecGeom single-precision proposal

agheata:
Added VecGeom proposal and removed the `John Doe` entry in the mentor list.

### DIFF
--- a/gsoc/2021/mentors.md
+++ b/gsoc/2021/mentors.md
@@ -3,12 +3,14 @@ title: Summary of Mentors GSoC 2021
 layout: plain
 ---
 
+**Note for contributors:** entries must be sorted in last name alphabetic order
+
 ## Full Mentor List (Name, Email, Org)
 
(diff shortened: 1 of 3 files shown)


gsoc/2021/mentors.md
```diff
@@ -3,12 +3,14 @@ title: Summary of Mentors GSoC 2021
 layout: plain
 ---
 
-## Full Mentor List (Name, Email, Org)
+**Note for contributors:** entries must be sorted in last name alphabetic order
```
xvallspl, line 6: Shouldn't this go in its own commit PR?

xvallspl (approved):

gsoc/2021/mentors.md
agheata: Yes in principle, but given that it is a minor detail changing the same file, I would let it like this for convenience.

agheata:
@graeme-a-stewart the link checker seems to have problem swallowing https://geant4.web.cern.ch is it because of https?

PR #944, opened 2021-03-15
fixes for Madgraph5

agheata:


agheata:
The folders I've made previously for withdrawn projects made still visible the MAdgraph5 proposal, now it should be fine. I also removed UC Louvain as org since there are no more projects attached.

xvallspl (approved):

PR #963, opened 2021-04-30
New curriculum design

klieret:
This not only looks better, it should also be much better to maintain. All the information is put into a YAML file. A selection of modules can then be shown with a simple include statement, selecting the relevant modules by their IDs.


klieret:
This should be approved by the training convenors, e.g. @meevans1

meevans1 (approved):
Looks great! Thanks @klieret and Claire!

PR #970, opened 2021-05-17
Generator WG meeting #12 minutes 06 May 2021 

valassi:


valassi:
Self merging

hegner:
Hi Andrea. Thanks for posting. I do not like self merging. A second pair of eyes is always useful.

(e.g. self-merging of minutes leads to the situation that minutes now cannot be found on the respective working group page itself as people did not do any tagging to allow this. Now it became a bigger task that has to be fixed for many, many meetings)

PR #976, opened 2021-06-02
Bugfix/baseurl troubles

klieret:
If you've ever tried to get github pages build our page from a fork, you will have noticed that a lot of things break. This is because of trouble with base urls: When you include something as `/myresource.png`, this will resolve to `www.mydomain.xyz/myresource.png`, even if your baseurl is set to `www.mydomain.xyz/subfolder`. This is what breaks the default github pages build, because it has the baseurl e.g. `https://klieret.github.io/hsf.github.io/`. 

In this PR I try to fix a lot of these occurrences. The last commit is already a month ago and I'm not sure if I have still overlooked some examples though. 

In any case this is something to watch out for in the future as well: Don't let people link as `[training](/training)` always do `[training]({{ site.baseurl }}/training)`.


graeme-a-stewart:
Hi @klieret - this look pretty nice and we could even make this improvement incrementally (we'll probably miss sime URLs first time around). We also need to update instructions as well, to give people good examples of how to do this.

klieret:
I resolved the conflicts. Let's merge this soon, because resolving these conflicts is a bit annoying. I am on vacation now, but I could update the instructions and give a quick update in the meeting in two weeks.

Actually the best solution to enforce this would be to write a very simple pre-commit hook.

My scripts basically just searched for things like `href="/`, so we might just check for something like that.

klieret:
I actually just put something together which seems to work really well: https://github.com/klieret/jekyll-relative-url-hook

klieret:
What do you think @graeme-a-stewart @hegner

hegner:
Looks very good. And like quite some tedious work to get to here,

hegner:
Can you try to investigate the failing tests though?

klieret:
I did most of this automatically. The failures are mostly known, but I saw one or two that I might just fix while we're at it. But they will never pass in total

klieret:
So I fixed some more links, the rest are old failures that require more work

hegner:
Thank you. Once the tests finished I'll merge

hegner (approved):

PR #980, opened 2021-06-25
Update training-modules.yaml

936-BCruz:
Adding the UNIX Shell training

### DIFF
--- a/_data/training-modules.yaml
+++ b/_data/training-modules.yaml
@@ -165,9 +165,9 @@
   webpage: https://hsf-training.github.io/hsf-training-ssh-webpage/
   id: ssh
 - description: A guide through the basics of the file systems and the shell.
-  name: UNIX
+  name: The UNIX Shell
   repository: https://github.com/swcarpentry/shell-novice
-  status: alpha
+  status: stable
   videos: ''
   webpage: https://swcarpentry.github.io/shell-novice/
   id: unix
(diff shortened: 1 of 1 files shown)


_data/training-modules.yaml
```diff
@@ -164,3 +164,10 @@
   videos: ''
   webpage: https://hsf-training.github.io/hsf-training-ssh-webpage/
   id: ssh
+- description: A guide through the basics of the file systems and the shell.
+  name: UNIX
+  repository: https://github.com/swcarpentry/shell-novice
+  status: alpha
```
klieret, line 170: this should be `stable`

_data/training-modules.yaml
```diff
@@ -164,3 +164,10 @@
   videos: ''
   webpage: https://hsf-training.github.io/hsf-training-ssh-webpage/
   id: ssh
+- description: A guide through the basics of the file systems and the shell.
+  name: UNIX
```
klieret, line 168: I think `The Unix Shell` would be better here. This is what becomes the title, and unix is a biiiiig topic ;)

klieret (changes requested):

klieret:
Hi @936-BCruz Thanks for starting on this right away ;)  You can address the two change requests by pushing more commits to your branch (not sure if you can do it from the web)

_data/training-modules.yaml
936-BCruz: You mean replacing _status: alpha_, with _status: stable_?

klieret (approved):

PR #989, opened 2021-07-01
Use gravatar for placeholder images if neither gravatar/github given

klieret:
If a person specifies neigher gravatar nor github we will still request an image from gravatar because it returns an empty picture

image


klieret:
Hi @hegner @graeme-a-stewart  could you merge this one and then [this one](https://github.com/HSF/hsf.github.io/pull/990)?

meevans1 (approved):

PR #990, opened 2021-07-01
New profiles

klieret:
Those are actually just 3 new profiles, but a lot more files were modified, because I was testing the new updater from https://github.com/HSF/website-helpers (which reads the results of a google survey) and it rewrote the corresponding files

### DIFF
--- a/_includes/profile_avatar.html
+++ b/_includes/profile_avatar.html
@@ -1,7 +1,7 @@
 <!-- Make sure to include educator_header to use this -->
 
-{% if person.gravatar %}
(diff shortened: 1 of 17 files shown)


klieret:
Whoops, actually this branch is already branched of https://github.com/HSF/hsf.github.io/pull/989

meevans1:
@klieret I'm not sure I can approve this one. Is it because of the failing checks?

klieret:
> @klieret I'm not sure I can approve this one. Is it because of the failing checks?

Hmm, maybe because you hadn't joined as a reviewer? Can you try again (I added you)

It's the link checker that makes the build fail. I will write to people to fix their homepage URLs, but I think this shouldn't block this

PR #991, opened 2021-07-01
feat: Update Giordon's profile (image and description)

kratsg:


kratsg:
Note: the `gravatar` doesn't seem to be working well. So hopefully we can rely on `https://github.com/kratsg.png` for this instead.

klieret (approved):

klieret:
@kratsg Thanks for the PR! Hmm, I can see the gravatar image but in either way with your change it should take the picture from github instead :)

PR #995, opened 2021-07-18
Training home improvements

klieret:
More pictures, more big links


klieret:
Probably best discussed during training meeting today

meevans1 (approved):

PR #998, opened 2021-07-19
Change sorting order of schools; add comments on top

klieret:


klieret:
This is a change that Sudhir requested to make it easier for him to add new events. Nothing is changed but the order of the entries

meevans1 (approved):

PR #1004, opened 2021-07-29
Add training resources

klieret:


_training/curriculum_table.md
```diff
@@ -2,89 +2,55 @@
 title: "HSF Training Modules"
 layout: plain
 ---
-
```
klieret, line 5: These changes in this file look complicated, but they aren't: All the css is moved to `hsf.css` and the table is placed in an additional `<div class="training">`.

michmx (approved):

PR #1014, opened 2021-09-08
Add codespell hook

klieret:
This hook checks for common spelling mistakes. False positives are very rare and can be excluded in `codespell.txt`


klieret:
I'll add documentation with a separate PR

klieret:
The remaining link failures look like the usual lot to me.

hegner (approved):
thanks a lot! :-)

PR #1017, opened 2021-09-14
Fix more absolute urls

klieret:
The pre-commit config pointed to a fixed and buggy version of the relative URL checker, so some absolute URLs were overlooked


klieret:
looks like our link checker doesn't like `{{site.baseurl}}`, but this can probably be fixed by editing [the config file](https://github.com/HSF/hsf.github.io/blob/master/.github/config/mdcheck.json). I'll open a separate PR for that

hegner (approved):

PR #1018, opened 2021-09-16
Add link checker replacement pattern for site.baseurl

klieret:
Currently the link checker complains about URLs like `{{ site.baseurl }} / get_involved.html`, because it doesn't expand `{{ site.baseurl }}`. Let's see if we can fix this


klieret:
I checked that this is working as expected in #1019 (note that some of the links there were supposed to be broken)

hegner (approved):
Great! :-)

PR #1020, opened 2021-09-16
pre-commit hook to check for correct syntax in JSON/YAML

klieret:
After seeing that the link checker silently ignored part of its config file due to wrong JSON syntax in #1018, this should always warn about issues with our json/yaml files.


hegner (approved):
Thanks Kilian! You really do it thoroughly

PR #1034, opened 2021-12-08
Add clarity on WG diversity

graeme-a-stewart:
We didn't say anywhere that converers from the same experiment is strongly discouraged, although we have always had this as a unwritten policy - so better that it is publically written down.

### DIFF
--- a/what_are_WGs.md
+++ b/what_are_WGs.md
@@ -19,7 +19,7 @@ agreement. (In contrast, [interest groups and activity areas]({{ site.baseurl }}
 are less formal.)
 
 We strive for a diverse representation in the HSF working groups and, in
-particular, we very strongly favour a group of conveners from three different
+particular, we very strongly favour having conveners from three different
 experiments.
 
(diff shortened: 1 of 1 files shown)


what_are_WGs.md
```diff
@@ -18,6 +18,10 @@ committee) to serve for one year, with renewal possible by mutual
 agreement. (In contrast, [interest groups and activity areas]({{ site.baseurl }}/what_are_activities.html)
 are less formal.)
 
+We strive for a diverse representation in the HSF working group and, in
```
hegner, line 21: "the groups" or "a group" ?

hegner (changes requested):

what_are_WGs.md
graeme-a-stewart: Yep, "groups" is better, thanks.

graeme-a-stewart:
Ping on this - I think it's good to go now.

hegner (approved):

PR #1036, opened 2021-12-10
Talks update

graeme-a-stewart:
These are all the external HSF talks that I could find in 2021

### DIFF
--- a/organization/presentations.md
+++ b/organization/presentations.md
@@ -1,9 +1,7 @@
 ---
 title: "HSF Presentations"
-layout: default
+layout: plain
 ---
-
-# HSF Presentations
 
 ## 2021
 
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues (approved):
Wow, quite a list. Kudos for the effort!

No need to go and check links since we have automated tests in place 👍.

All look good to me.

organization/presentations.md
```diff
@@ -1,16 +1,19 @@
...
 title: "HSF Presentations"
-layout: default
+layout: plain
 ---
 
-# HSF Presentations
-
 ## 2021
 
+* [HEP Computing Challenges and the HSF](https://indico.physics.lbl.gov/event/1756/contributions/6247/), [Snowmass Small Experiments Workshop](https://indico.physics.lbl.gov/event/1756/), 16 November 2021, Elizabeth Sexton-Kennedy
```
eduardo-rodrigues, line 8: Maybe just the comment that in this case you did not add the direct link to the PDF?

graeme-a-stewart: Right - which is what I tend to do when you can link to the "contribution", so it's deliberate. The direct slides link can be fragile if the slides get updated and the URL changes (albeit unlikely post facto, but still...).

Issue #1046, opened 2022-01-18
Link checker fails internal links


klieret:
The link checker fails all links in recent PRs, see e.g. https://github.com/HSF/hsf.github.io/pull/1045 :+1: 

> All the link check failures seem to be the same kind as yesterday and all false positives. Maybe @klieret has an idea - it seems that {{site.baseurl}} is expanding to an empty string and then the URL checker is trying for an absolute path, instead of a relative one.

klieret:
Hmm, I have no idea why this isn't working.

I confirmed with a `find` statement for example for the existence of

```
/__w/hsf.github.io/hsf.github.io/_site/training/curriculum.html
```

then later the link checker complains that

```
2022-01-18T17:37:10.3403364Z [✖] /__w/hsf.github.io/hsf.github.io/_site/training/curriculum.html → Status: 400 [Error: ENOENT: no such file or directory, access '/__w/hsf.github.io/hsf.github.io/_site/training/curriculum.html'] {
2022-01-18T17:37:10.3403782Z   errno: -2,
2022-01-18T17:37:10.3404010Z   code: 'ENOENT',
2022-01-18T17:37:10.3404232Z   syscall: 'access',
2022-01-18T17:37:10.3404597Z   path: '/__w/hsf.github.io/hsf.github.io/_site/training/curriculum.html'
2022-01-18T17:37:10.3404854Z }
```

klieret:
I'm currently out of ideas @graeme-a-stewart 

What we could do as a not-perfect solution is to use the replacement pattern to prefix `https://hepsoftwarefoundation.org`. This would fail if you add a new markdown file and then directly link to it in the same PR, but should work for all the other links.

graeme-a-stewart:
Just spitballing here, would it help if instead of 

```json
    "pattern": "^\\s*{{\\s*site.baseurl\\s*}}/(.*)",
    "replacement": "/_site/$1"
```

we made it a relative path, viz.

```json
    "pattern": "^\\s*{{\\s*site.baseurl\\s*}}/(.*)",
    "replacement": "./_site/$1"
```

klieret:
Nope, that also doesn't work:

```
[✖] _site/training/curriculum.html → Status: 400 [Error: ENOENT: no such file or directory, access '/github/workspace/_workinggroups/_site/training/curriculum.html'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'access',
[...]

PR #1051, opened 2022-01-21
Fix link checker for most cases

klieret:
This is only a half-hearted fix: It will fail if you create a new page
and link to it before it is published.
See https://github.com/HSF/hsf.github.io/issues/1046 for more
information.


klieret:
(all internal links pass now, the remaining failure seems to be legit)

klieret:
(currently rerunning the last check)

hegner (approved):
ok. seems like a little temporary hack. ok with me for now. thanks

PR #1056, opened 2022-02-02
Add Conditions Database activity page, _activities/conditionsdb.md

paullaycock:
Adding the Conditions Database activity page, first draft (checked with local jekyll installation)

### DIFF
--- a/_activities/conditionsdb.md
+++ b/_activities/conditionsdb.md
@@ -1,0 +1,36 @@
+---
+title: "Conditions Databases"
+author: "Paul James Laycock"
+layout: plain
+---
+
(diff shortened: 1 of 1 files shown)


hegner (changes requested):
Thanks @paullaycock for adding it.
Would be great if you could set up a mailing list and then add it to this page. Then I am happy to merge

paullaycock:
Hi @hegner , I was going to do that (and also a coordination list), it's like a reflex, but @graeme-a-stewart pointed out that it's unlikely the list would be used for productive discussion (at least I wouldn't want to use an email list for that) and we'd probably like the few meeting announcements to be on the main forum, not least given the connections with other working groups.  Maybe a short point for discussion in the coordination meeting today, but I found the argument compelling!

hegner:
@paullaycock - I think you (and Graeme) have a point there.

hegner (approved):

PR #1058, opened 2022-02-07
WIP: Gsoc root ttree to rntuple proposal

aflp91:


hegner (approved):

agheata:
I don't see the difficulty, duration and availability labels in this PR. We need to put the in place so they get visible in the final proposal page.

aflp91:
Yes ! And it is why the PR was WIP!

hegner:
ops. sorry!

hegner:
I took it out again.

agheata:
No problem, we can do this in a separate patch. But we'll need the logic to maybe transform these labels into text just below the title of the project proposal

hegner:
That would be by changing 
https://github.com/HSF/hsf.github.io/blob/master/_layouts/gsoc_proposal.html
and using tags like {{ page.duration }}
Anyone of you signing up for that? Otherwise I will do this afternoon

agheata:
Working on a couple of other items, and my afternoon is busy I'm afraid

aflp91:
Another possibility is to add in `summary.md` and under the `Full list of proposals` lists by duration, by difficulties and by availability?

agheata:
We may also do that to give a global overview (good idea), but this should appear also in the proposals for self-consistency

hegner:
@jalopezg-r00t - thanks a lot. exactly what was needed
will review

PR #1074, opened 2022-02-14
Striped design for alpha/beta modules

### LINKED ISSUE #1073: Colorblindness-friendlness on training curriculum page (for discussion)
The colors on the [training curriculum page](https://hepsoftwarefoundation.org/training/curriculum.html) are close together in intensity, and use a color scheme that is not very suitable to people with colorblindness. Although the same information is already present in the text (Status: Early development, Status: Beta testing), it may make sense to adopt a visual cue that is independent of color as well.

I played around a bit with repeating gradients to achieve this within the existing color scheme, and came up with this:
image

The additional modality is the fraction of the rectangle that is filled by foreground color. The 45 degree angled stripes are to hint at e.g. police caution tape.

Here is the relevant css ([browser compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-linear-gradient()#browser_compatibility)):
```css
.tm-alpha {
  background-image: repeating-linear-gradient(-45deg, transparent, transparent 20px, #f9f2f4 20px, #f9f2f4 40px);
}
.tm-beta {
  background-image: repeating-linear-gradient(-45deg, transparent, transparent 10px, #f9f9f1 10px, #f9f9f1 40px);
}
```

I feel that this now distracts a bit too much, but I think a similar approach limited to the border of the rectangles could be effective without adding too much visual clutter.

klieret:
Meant to make the curriculum more color-blind friendly.

As proposed and implemented by @wdconinc.

Closes #1073

### DIFF
--- a/css/hsf.css
+++ b/css/hsf.css
@@ -292,11 +292,10 @@ body {
 }
 
 .tm-alpha {
-  background-color: #f9f2f4;
+  background-image: repeating-linear-gradient(-45deg, transparent, transparent 20px, #f9f2f4 20px, #f9f2f4 40px);
 }
-
 .tm-beta {
-  background-color: #f9f9f1;
+  background-image: repeating-linear-gradient(-45deg, transparent, transparent 10px, #f9f9f1 10px, #f9f9f1 40px);
 }
 
 .training-module-name {


klieret:
image

hegner (approved):

PR #1080, opened 2022-02-16
Fix a wrong space in MadAnalysis 5 project GSoC proposal

aflp91:
Rename the MadAnalysis project file to be consistent with the project name

### DIFF
--- a/_gsocproposals/2022/proposal_MadAnalysis-mcweights.md
+++ b/_gsocproposals/2022/proposal_MadAnalysis-mcweights.md
@@ -1,7 +1,7 @@
 ---
 title: MadAnalysis 5 - Integration of theoretical uncertainty calculation with multiweight integration
 layout: gsoc_proposal
-project: MadAnalysis 5
+project: MadAnalysis5
 year: 2022
 difficulty: medium
 duration: 350


agheata:
@aflp91 This makes the proposal not to be correctly associated with the project (while it currently works correctly). The project name can have more words, what is important is the matching in the proposal.

_gsocproposals/2022/proposal_MadAnalysis-mcweights.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: MadAnalysis 5 - Integration of theoretical uncertainty calculation with multiweight integration
 layout: gsoc_proposal
-project: MadAnalysis 5
+project: MadAnalysis5
```
agheata, line 4: This does not work in my test, the MadAnalysis project p[age shows no proposals. This works without this change.

agheata (changes requested):

_gsocproposals/2022/proposal_MadAnalysis-mcweights.md
aflp91: That's the opposite on my local site ! I missed something…

agheata: Well, the best check is the regular project page which renders correctly as it is now: https://hepsoftwarefoundation.org/gsoc/projects/2022/project_MadAnalysis5.html

aflp91: Andrei, your are saying the link following `Corresponding Project` in `proposal_MadAnalysis-mcweights.md` is working without renaming `project_MadAnalysis5.md` to `project_MadAnalysis 5.md` ?

agheata: No, just the link from the project page to the proposal, which disappears with your patch. But indeed, the back-link from the proposal to the project is not working. So probably you are right to remove the space, only that we should do it also in the project file `project_MadAnalysis5.md` by using `project: MadAnalysis5`

agheata (approved):

PR #1082, opened 2022-02-17
Fix link from MA5 project to its proposal.

agheata:
The to MadAnalysis 5 project page has the link to the proposal with this fix.


aflp91 (approved):
Looks OK with me!

PR #1103, opened 2022-02-22
Fix Escape image suffix

aflp91:
@agheata, @hegner : the `https://hepsoftwarefoundation.org/gsoc/organizations/2022/escape.html` page is empty and  right now I don't see why … so the WIP …


_gsocorgs/2022/escape.md
```diff
@@ -3,7 +3,7 @@ title: "ESCAPE"
 author: "Enrique Garcia"
 layout: default
 organization: ESCAPE
-logo: Escape-logo.jpg
```
garciagenrique, line 6: good catch ! thanks !

PR #1104, opened 2022-02-25
GSoC proposal for Julia/PODIO

hegner:


_gsocproposals/2022/proposal_PODIO-Julia,md
```diff
@@ -0,0 +1,90 @@
+---
```
tmadlener, line 1: The filename seems to have a `,` (comma) instead of a `.` (dot) before the `md`. This might break githubs automatic rendering.

_gsocproposals/2022/proposal_PODIO-Julia,md
```diff
@@ -0,0 +1,90 @@
...
+##  PODIO
+
+PODIO is a data-model library based on the idea of representing information as plain-old-data (PODs) 
+instead of using a deep object-oriented model. This allows for better performance and interoperability.
+Data models are defined by a dedicated YAML-syntax and the corresponding code is auto-generated. So far 
+both C++ and Python are supported.
+
+## Task ideas and expected results
+
+The task is to create a code generator to add support for Julia as end-user language, and to carry out
```
tmadlener, line 41: Maybe we could specify that "the code generator" in this case would involve mainly some new jinja templates as the main infrastructure for code generation is already in place. But for the project proposal that might be too detailed. I am also fine with the current formulation.

tmadlener (commented):
Thanks for preparing this. Just two minor comments from my side, see inline.

_gsocproposals/2022/proposal_PODIO-Julia,md
hegner: thanks. I think we can leave it as is for now.

hegner:
@agheata - it is sitting here for a few days now... can you check?

agheata (approved):

agheata:
Note that there is a codespell error: _gsocproposals/2022/proposal_PODIO-Julia.md:52: representaiton ==> representation
I merged nevertheless but you are welcome to fix it

PR #1113, opened 2022-03-08
Update Gemfile

graeme-a-stewart:
A 'fresh' checkout of the website on a new Windows machine, which also had Ruby 3.1.1,
required that the removal of the platform specifier for tzinfo-data and an explicit addition of webrick.

These changes were backported to an older setup (Ruby 2.7.2 on OS X)
and they don't cause any problems there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,8 @@ gem "html-proofer"
 # Suggested by Jekyll
 gem 'wdm', '>= 0.1.0' if Gem.win_platform?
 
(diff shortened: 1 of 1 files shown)


graeme-a-stewart:
Ugh, well while ok for Ruby 2.7.2, it's not OK for Ruby 2.6.0, which apparently is what our CI is doing. I will take a look at that.

PR changed to Draft for now.

graeme-a-stewart:
Using the better `if Gem.win_platform?` seems to avoid problems

PR #1126, opened 2022-04-01
Correct a wrong example in GSoC proposal.

junaire:


graeme-a-stewart (approved):
Approving based on @vgvassilev's comment

PR #1135, opened 2022-04-26
Always check for case conflicts

klieret:


klieret:
Hi @hegner @graeme-a-stewart Could you approve?

graeme-a-stewart (approved):
LGTM - if I read this correctly then the CI failure is expected in this case (there is a case clash). So this should indeed be safe to merge, right @klieret?

klieret:
Exactly :)  Since my other PR solved the clash, this should now work fine on our main branch

PR #1155, opened 2022-06-13
Include matplotlib training in "basics"

klieret:


klieret:
You can ignore the failed pre-commit, that's due to another change and I'll fix it in a separate PR

michmx (approved):

PR #1156, opened 2022-07-04
Fix absolute URL

klieret:


eduardo-rodrigues (approved):
I trust you on this one. I had copied the example from somewhere else.

PR #1172, opened 2022-07-25
Add GSoC 2022 blog from Kush Kothari

kkothari2001:
Please have a look @jpivarski


jpivarski (approved):
It looks good to me! Maybe link to the [dask-awkward project](https://github.com/ContinuumIO/dask-awkward/) when you refer to it?

PR #1198, opened 2022-08-08
Adding Aman Desai to Training Community

klieret:


klieret:
Link checker failure will be solved by #1199

michmx (approved):

PR #1200, opened 2022-08-08
Matplotlib no longer is in early development

### LINKED ISSUE #1189: Matplotlib is mentioned twice on the curriculum page

klieret:
Fixes #1189


klieret:
Link checker failure will be solved by #1199

PR #1210, opened 2022-08-12
Order of badges is now fixed, so remove instruction

klieret:


hegner:
one of the link check failures addressed with #1211

michmx (approved):

PR #1213, opened 2022-08-14
Update profiles: educators w/ material

klieret:


hegner:
@klieret - seems it still is testing against the wrong version of Graeme's profile. Can you make a quick update of this PR?

klieret:
It should work now :)

hegner:
I do not like the term "material" there. Wouldn't "author" or "writer" be better? Since that is the role, not the result?

michmx (approved):
I also agree that it is a better idea to switch from "material" to "author". But I see this must be done in all profiles that include the role, so let's go ahead as Kilian propose and fix that in a different PR.

PR #1214, opened 2022-08-14
Use link blocks instead of hard-to-parse text

klieret:


hegner (approved):
Usually better not to put too much bare HTML into things, but OK with me in this case.

klieret:
Thanks for the review!

Is there any way to do something like this with markdown? There's the pandoc style class injection, but even that would probably not be enough (and I doubt that it'd be more stable).
I could conjure up something with liquid includes or so, but I doubt it'd be more pretty either.

But I'd be interested in the motivation of your recommendation: Why do you think that having bare HTML is a bad thing?

PR #1215, opened 2022-08-15
Update GraemeS training profile

graeme-a-stewart:
Just a years update!


klieret (changes requested):
The link to the profiles.web seems to be dead...

hegner (approved):

klieret:
We should still fix the link though @graeme-a-stewart

PR #1221, opened 2022-08-22
Add C++ school

klieret:


klieret:
Changes are a bit noisy because script enforces alphabetical order

michmx (approved):

PR #1224, opened 2022-08-29
Rename material badge to author badge

klieret:


klieret:
Link checker is false positive

hegner (approved):
failed check is indeed a glitch. 

Thanks for calling humans not material any more by mistake ;-)

PR #1225, opened 2022-08-29
Make sure {{ site.baseurl }} actually shows up in documentation

klieret:


klieret:
Took me a few minutes to figure out what was going on here, because the `site.baseurl` bit is so far just not shown on the rendered guide haha

hegner (approved):

PR #1230, opened 2022-09-02
Standardize name: curriculum -> training center

klieret:


hegner:
@klieret - is that something to be agreed with @michmx still or is it just the technical implementation of the change?

klieret:
Just technical/standardization/consistency :)

hegner (approved):

PR #1248, opened 2022-09-13
Remove sitemap; add twitter inset

klieret:
All of the links of the sitemap already appear in other places in a more sorted fashion, so we shouldn't waste so much prime real estate

### DIFF
--- a/_workinggroups/training.md
+++ b/_workinggroups/training.md
@@ -27,30 +27,11 @@ The group aims to develop a training program that can be pursued by researchers 
   </a>
 </div>
 
-## Sitemap
 
-* I'm a student and want to **learn** about HEP Software:
-  * [The HSF training center]({{site.baseurl}}/training/curriculum.html)
-  * [List of upcoming training events]({{site.baseurl}}/Schools/events.html)
-* I want to **teach** software:
-  * Training events:
-    * [How to organize a software training event]({{site.baseurl}}/training/howto-event.html)
-    * [The different roles in an HSF-training event]({{site.baseurl}}/training/educators.html)
-  * Training modules:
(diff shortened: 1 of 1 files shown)


klieret:
image

_workinggroups/training.md
```diff
@@ -27,30 +27,11 @@ The group aims to develop a training program that can be pursued by researchers
...
-    * [The HSF training community]({{site.baseurl}}/training/community.html)
-    * [Participating & Contributing]({{site.baseurl}}/workinggroups/training.html#how-to-participate-and-contribute)
-    * [Convenors]({{site.baseurl}}/workinggroups/training.html#conveners)
-* **Meta**:
-  * [How to add your profile to the HSF training community pages]({{site.baseurl}}/howto-profile.html)
+
+## News
+
+<a class="twitter-timeline" style="margin: auto" data-width="100%" data-height="500" href="https://twitter.com/HSFTraining?ref_src=twsrc%5Etfw">Tweets from @HSFTraining</a> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script> </div>
+
```
wdconinc, line 34: I worry 100% width breaks the page and people won't read beyond the twitter box. Is there a way to move it down to e.g. "Participate & Contribute" or "Training Events"?

klieret: We'd have to see how this plays out on mobile (which the whole webpage is unfortunately not really optimal for), but on desktop it's very clear.

Also, even if the thing with the two scroll bars is not entirely clear on mobile right away: The twitter feed is limited to ~5 entries, so scrolling there naturally stops and you resume scrolling the main page.

wdconinc (approved):

Issue #1247, opened 2022-09-13
Fix mobile layout of floating pictures


klieret:
* **Required knowledge**: HTML, CSS

The floating pictures on https://hepsoftwarefoundation.org/workinggroups/training.html get very very small on a mobile layout. Instead they should just flow to "interrupt" the text.

developer-irfan:
Hello @klieret, yes pictures get very small on mobile device bcz they are not responsive and are in the same row of text. just assign me this i will solve this so in order the pics are display properly. thankyou

klieret:
Thank you for your interest @developer-irfan !  There was already one PR (#1250) about this, but unfortunately I do not think that it does what we need. I'd be quite happy if you'd give it a try as well or leave a comment at #1250 :)

Ahnaf-codes:
is this good? @klieret 
![Screenshot 2022-10-06 at 22-40-55 HSF Training](https://user-images.githubusercontent.com/102795582/194370370-3dbe568a-406a-435f-87bb-4c59801a5bfc.png)

klieret:
Thanks for taking a look at that :) Already much better! But could we in principle also have it simply take 100% of the width if we are at such small display sizes?

Ahnaf-codes:
yeah thats pretty simple. ill send a pic

Ahnaf-codes:
![Screenshot 2022-10-06 at 22-56-24 HSF Training](https://user-images.githubusercontent.com/102795582/194373328-5c292bef-1814-49cc-a397-650454644a6f.png)
@klieret

klieret:
That looks perfect! :D

Ahnaf-codes:
ill send the pr tomorrow since im kinda tired

Ahnaf-codes:
@klieret its done

PR #1250, opened 2022-09-13
floating pictures on training.html is set to max-width:50%

Manoharvuradi:
By setting max-width to 50% gives correct size.

### DIFF
--- a/_workinggroups/training.md
+++ b/_workinggroups/training.md
@@ -9,7 +9,7 @@ redirect_from:
 excerpt: none
 ---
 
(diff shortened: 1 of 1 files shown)


klieret:
Hi @Manoharvuradi . Thanks for your PR in reference to #1247 

However, wouldn't this make the pictures significantly larger than they currently are on desktop? Also on mobile, 50% might still be too small, right?

klieret:
Again, thank you for your PR, but I don't think this is exactly achieving the goal of #1247. I will therefore close this for now, but feel free to open another PR. Just so you know, there was also interest by @developer-irfan to work on this (see #1247).

PR #1267, opened 2022-10-05
Replace pyhep announcement with hacktoberfest

klieret:


graeme-a-stewart (approved):
Perfect - I had been thinking that we needed an update here. Thanks @klieret!

PR #1268, opened 2022-10-05
docs: replace Hacktoberfest image

amangoel185:
Add a white background for better visibility in both light and dark mode, similar to  hsf-training/.github#11


klieret:
Thanks again :)

PR #1269, opened 2022-10-06
Made big link containers responsive

### LINKED ISSUE #1246: Fix mobile layout of "big links" used by training group
This is what it looks like on mobile atm:

We should probably switch to a `max-width` and flexible number of columns. Like employed on the main page of HSF

Ahnaf-codes:
Resolves #1246 by adding a media query breakpoint at 620px.

### DIFF
--- a/css/hsf.css
+++ b/css/hsf.css
@@ -372,3 +372,11 @@ body {
 .training table tr th :last-child, .training table tr td :last-child {
   margin-bottom: 0;
 }
+
+/* Media Queries  */
+
+@media (max-width: 620px) {
+  .big-link-container {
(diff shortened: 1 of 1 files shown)


klieret:
Awesome! I did not know about this at all :D 

But I'm wondering: Would it also be possible to reduce the height of the boxes when in single column mode? Currently it looks like this:

image

But only if there's a simple solution for this, I don't think this is worth to complicate the code much for :)

klieret:
Yes, exactly, this is perfect 🎉 ! Thanks a lot :) 

(and would you perhaps be interested in taking a look at #1247 ? ;) )

klieret (approved):

PR #1270, opened 2022-10-06
Add educator Jonas Rembser as 2022 instructor

guitargeek:
Thank you very much!

@klieret


klieret (approved):
Thanks a lot for adding yourself! :)

PR #1271, opened 2022-10-07
Fixed image inclusion for mobile layout

### LINKED ISSUE #1247: Fix mobile layout of floating pictures

Ahnaf-codes:
Resolves #1247 by making images responsive

### DIFF
--- a/_workinggroups/training.md
+++ b/_workinggroups/training.md
@@ -9,7 +9,7 @@ redirect_from:
 excerpt: none
 ---
 
-<img src="{{site.baseurl}}/images/training/analysis_preservation_bootcamp_participants.jpg" style="float: right; max-width: 30%; margin-left: 20px;">
+<img src="{{site.baseurl}}/images/training/analysis_preservation_bootcamp_participants.jpg" style="float: right; width: 90vw; max-width:320px; margin-left: 20px;margin-bottom: 20px;">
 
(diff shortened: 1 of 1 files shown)


klieret:
Thanks a lot for working on this! However, when I naively resize the browser window, I'm getting this: 

image

which makes the text appear very awkard (because only two words per line). Is there a way to solve this? Or will the phone display be ok? (the screenshot you showed in the issue seemed OK)

Ahnaf-codes:
Can you tell me what viewport width you're using? I can figure it out if I get the viewport width

klieret:
Thank you :)  Do you mean these values:

```
./_layouts/default.html:    <meta name="viewport" content="width=device-width, initial-scale=1 user-scalable=no">
./_layouts/meetings.html:    <meta name="viewport" content="width=device-width, initial-scale=1 user-scalable=no">
```

Ahnaf-codes:
No I meant the size of the screen. It's okay if you don't have it. But it may take a bit longer

klieret:
But don't different phones have different screen sizes?

Ahnaf-codes:
Yeah but I can resize my desktop window in responsive mode to see what the issue is.
So if I got an exact resolution I wouldn't have to play around with the resolution to get a good size

klieret:
I think the viewport above is around 495px wide. And when I just checked on my 6" phone, the viewport was still smaller, I think.

klieret:
Since this is definitely an improvement, let me go ahead and merge this!

Thanks again for your contributions @Ahnaf-codes

klieret (approved):

PR #1275, opened 2022-10-12
update link cpp training

klieret:
- Linking to links in readme rather than outdated
- Drop and move modules
- Make scikit-HEP stable

### DIFF
--- a/_training/curriculum.md
+++ b/_training/curriculum.md
@@ -21,7 +21,7 @@ The curriculum is comprised of a set of standardized *modules*, so that students
 
 ### C++ corner
 
-{% include list_of_selected_training_modules.html ids="hepcpp,moderncpp,cmake"%}
+{% include list_of_selected_training_modules.html ids="hepcpp,cmake"%}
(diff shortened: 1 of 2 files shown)


klieret:
Jim has informed me that we can move `uproot` to the archived section.

_training/curriculum.md
```diff
@@ -21,23 +21,23 @@ The curriculum is comprised of a set of standardized *modules*, so that students
 
 ### C++ corner
 
-{% include list_of_selected_training_modules.html ids="hepcpp,moderncpp,cmake"%}
+{% include list_of_selected_training_modules.html ids="hepcpp,cmake"%}
```
wdconinc, line 24: Should we add moderncpp to archived too then?

klieret: 👍  Ah sure. Though there is really not much to be seen there.

wdconinc (approved):

klieret:
Thanks!

PR #1278, opened 2022-10-21
Add QR codes for HSF and PyHEP

eduardo-rodrigues:
It would be ideal to have somewhere a sentence saying we have this material for presentations but I haven't found the perfect place for it. Ideas?


hegner (approved):

hegner:
I think we can just put it onto our next presentations ourselves to spread it. On the website I would not know a good place - maybe the get involved page?

eduardo-rodrigues:
An alternative would be to add a link at https://hepsoftwarefoundation.org/organization/presentations.html to a page on resources for presentations. Let's see how things go in the future and if we have more resources for presentations.

hegner:
That's a good idea. Actually we could as well link the usual template there

eduardo-rodrigues:
> That's a good idea. Actually we could as well link the usual template there

OK, to wrap the week I tried to give this a go, see PR https://github.com/HSF/hsf.github.io/pull/1279 :).

PR #1279, opened 2022-10-21
New page for presentation materials

eduardo-rodrigues:
So far only contains links for the QR codes I added in PR https://github.com/HSF/hsf.github.io/pull/1278.


hegner (approved):
Looks good to me!

PR #1287, opened 2022-11-07
Add presentations from 2022

graeme-a-stewart:
We should add this as a news item (this week) for people to update if they did any HSF talks this year. At least this is a start...


hegner (changes requested):
Can you switch to reverse chronological order? As was done before.

hegner (approved):
thanks

PR #1305, opened 2022-11-30
Add training statistics

klieret:
image


wdconinc (approved):
Looks very nice. Do we at some point need an 'Analysis' or 'Impact' page?

klieret:
I like the idea of a dedicated page! I also want to make these diagrams interactive (e.g., with `plotly`) in the future

PR #1313, opened 2023-01-09
Bump years of some educators

klieret:


wdconinc (approved):
Looks fine to me.

It seems that we encourage some consistency in yaml list syntax, but it's just the original entries that are multi-line lists, right? No need to change though as long as the template has a specific suggestion.

klieret:
We could just do a read-in/write out at some point (though it would remove the comments)... I had already built something like that here github.com/hsf/website-helpers/

PR #1315, opened 2023-01-09
fix: ignore failing link check of belle2.jp

wdconinc:
Per https://github.com/gaurav-nelson/github-action-markdown-link-check.


klieret (approved):

klieret:
Thanks!

Issue #1260, opened 2022-09-26
Jekyll: Fix superfluous closing div in list of educators per year


klieret:
**Required knowledge:** Jekyll, liquid

The logic that needs to be fixed is at https://github.com/HSF/hsf.github.io/blob/master/_includes/list_of_educators.html

If a new year starts and now educator has been registered for that year, a superfluous `</div>` is printed (see also comment at the bottom of the page)

himanshu007-creator:
Hi @klieret , i would like to work on this issue, can you help me understand what needs to be done?

Bikisome:
hello @klieret sir i am a begginer in this open source community .i would like to contribute to this issu can you please help me about this .

satyamsaxena2001:
![Screenshot from 2023-01-20 22-16-58](https://user-images.githubusercontent.com/95117179/213755723-92614aa4-9dcd-493e-965e-9701b40a7bf4.png)

@klieret is this the right way to do it. I would like to work on this if you can assign me the issue.

klieret:
@satyamsaxena2001 Looks good! Feel free to create a PR with this solution!

satyamsaxena2001:
@klieret i have created a PR.Please have a look and let me know.

Chiranjeev-Kartik:
@klieret  Since the issue is still open, I would like to work on it.

Xiexus26:
I would like to work on this issue since it is still open.

PR #1326, opened 2023-01-20
Fix superfluous div in new year for list of educators

satyamsaxena2001:
#1260 
To fix this bug, I added a check at the end of the inner loop (the one that loops through the educators) to see if any educators have been added to the current year. To accomplish this I added a variable, educators_added, that is set to false at the start of the inner loop, and then set to true whenever an educator is added to the current year. At the end of the inner loop, I check if educators_added is true and if so, added the closing div. If not, skipped the closing div.

I also added a else statement inside the unless loop that checks for the year, where I added the line to set the variable educators_added to true.

I also added a unless statement which checks whether the last column was full or not and also whether any educators were added in that year or not. If the column was not full and no educators were added, then it skips the closing div.

### DIFF
--- a/_includes/list_of_educators.html
+++ b/_includes/list_of_educators.html
@@ -14,11 +14,14 @@ github.com.com/carpentries/carpentries.org/gh-pages/_includes/community_profiles
 We need to keep track of how many educators are skipped in order to have the
 proper index for the table column. -->
 {% assign skipped = 0 %}
+{% assign educators_added = false %}
 {% for person in site.profiles| sort:"title" %}
 {% unless person.training_years contains loopyear %}
 {% assign skipped = skipped | plus: -1 %}
 {% continue %}
-{% endunless %}
+{% else %}
+{% assign educators_added = true %}
+{% endif %}
 
 {% assign column = forloop.index | plus: skipped | modulo: 4 %}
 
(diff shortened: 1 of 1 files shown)


_includes/list_of_educators.html
```diff
@@ -14,11 +14,14 @@ <h3 id="{{ loopyear }}"> {{ loopyear }} </h3>
...
 {% assign skipped = 0 %}
+{% assign educators_added = false %}
 {% for person in site.profiles| sort:"title" %}
 {% unless person.training_years contains loopyear %}
 {% assign skipped = skipped | plus: -1 %}
 {% continue %}
-{% endunless %}
+{% else %}
+{% assign educators_added = true %}
+{% endif %}
```
klieret, line 24: shouldn't that be an `endunless`?

klieret (changes requested):

_includes/list_of_educators.html
satyamsaxena2001: Added

klieret: Sorry, I meant that the `endif` should be an `endunless`, not to add another `endunless` before

klieret (changes requested):

_includes/list_of_educators.html
satyamsaxena2001: done

klieret (approved):
Thanks a lot again!

PR #1329, opened 2023-01-31
Add MCnet project, UGlasgow org, and LHAPDF proposal

agbuckley:
A new proposal for LHAPDF work, re-enabling MCnet and UGlasgow to support.


hegner (approved):
Thanks @agbuckley

PR #1330, opened 2023-02-03
Implemented conditional display of school list title based on emptiness.

Chiranjeev-Kartik:
**Problem statement:** https://hepsoftwarefoundation.org/Schools/events.html: The current/upcoming section is shown even without corresponding events. **_Issue: #1298_**
To ensure that the title is only displayed when the list of upcoming schools in "list_of_upcoming_schools.md" is not empty, a conditional statement is added in the "events.md" file. The statement checks the emptiness of the list and only displays the title if it is not empty.

### DIFF
--- a/Schools/events.md
+++ b/Schools/events.md
@@ -3,9 +3,12 @@ layout: plain
 title: HEP Software Training Events
 ---
 
-## Current and Upcoming Training Events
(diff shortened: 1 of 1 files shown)


klieret (approved):
LGTM! Thanks :)

PR #1331, opened 2023-02-03
Add cpp course announcement

klieret:


wdconinc (approved):
Looks good. Is there a use for a tag `virtual`, `hybrid`, `in-person`? Not sure how tags are defined and if we can just add them by text.

klieret:
> Looks good. Is there a use for a tag virtual, hybrid, in-person? Not sure how tags are defined and if we can just add them by text.

Good idea, I've opened https://github.com/HSF/hsf.github.io/issues/1343 for it

PR #1332, opened 2023-02-03
Fix test for empty workshop list

klieret:
Fixes https://github.com/HSF/hsf.github.io/pull/1330, this time tested locally

(also copies solution over to training WG home)


wdconinc (approved):
Looks good to me, but I didn't test locally.

PR #1336, opened 2023-02-06
Add proposal for SWAN SparkConnect project

etejedor:
Adds proposal for SWAN SparkConnect project and the corresponding info for the SWAN project.


hegner (changes requested):
Thanks. Can you update https://github.com/HSF/hsf.github.io/blob/master/gsoc/2023/mentors.md ?

PR #1337, opened 2023-02-06
Alpaka CLUEstering - Integrate Alpaka in CLUE python package

waredjeb:
Add proposal for Patatrack project, Alpaka CLUEstering

PR #1339, opened 2023-02-06
patatrack gsoc proposal: fix missing whitespace in yaml

vvolkl:
https://github.com/HSF/hsf.github.io/pull/1337 does not render correctly at the moment at https://hepsoftwarefoundation.org/activities/gsoc.html - this should fix it.


hegner:
Thanks for spotting

PR #1350, opened 2023-02-07
more gsoc jekyll fixes

vvolkl:


hegner:
@vvolkl - quicker than me sometimes :-D

PR #1355, opened 2023-02-07
Fix codespell ignore list

klieret:
Fixing ineffective #1349. The ignore-listed words must always be lowercase AFAIK


hegner:
Thanks

PR #1357, opened 2023-02-07
Fix wonky indentation

klieret:
Not sure why, but the whitespace caused strange indents to show up on the rendered page

image


klieret:
pre-commit fail is unrelated and will be fixed in #1355

PR #1369, opened 2023-02-14
Absolute paths for link checker and baseURL

klieret:
In reference to #1046

### DIFF
--- a/.github/config/mdcheck.json
+++ b/.github/config/mdcheck.json
@@ -19,13 +19,14 @@
   "replacementPatterns": [
     {
       "pattern": "^/(.*)",
-      "replacement": "https://hepsoftwarefoundation.org/$1"
+      "replacement": "/_site/$1"
     },
     {
       "pattern": "^\\s*{{\\s*site.baseurl\\s*}}\\s*/(.*)",
-      "replacement": "https://hepsoftwarefoundation.org/$1"
+      "replacement": "/_site/$1"
     }
   ],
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues:
Hello @klieret, cc @graeme-a-stewart and @hegner for website enhancements. I see this long-standing PR. It's rather short. Do you intend to finalise or else we close at this point?

eduardo-rodrigues:
Hello @klieret, what is the status here? Let me also ping @graeme-a-stewart and @hegner since they are thinking about the new revamped site.

jouvin:
Is this PR still relevant? I have the feeling that yes and it would be a good idea not to hardcode the host name of the web site? If yes, why we don't merge this PR is @klieret is not available to follow it up?

eduardo-rodrigues:
Probably. I'm no expert but the changes seem small eough that somebody can review?

graeme-a-stewart:
I can take a look, but I would say this wants a rebase first!

graeme-a-stewart:
OK, this all looks good to me. Killian had set it to link check the whole site, but that took about an hour and we used to get badly burned by false positives when that was done on PRs. So I have reverted to the check of changed pages only. (It would be good to also have a workflow that triggers a whole site check, but that shouldn't be the normal workflow for PRs.)

graeme-a-stewart (approved):
Thanks a lot for this change @klieret!

PR #1379, opened 2023-02-24
Applicants should include ML

klieret:


wdconinc (approved):
Looks good to me. Maybe add "for this project" because you might get cc'ed on some other projects too (never underestimate the ability of applicants to extrapolate).

wdconinc (approved):

PR #1381, opened 2023-02-26
Project key must be same as filename part

klieret:
Seems like backlink from proposal to project depends on this congruence


wdconinc (approved):
LGTM

PR #1389, opened 2023-03-05
chore: cleanup `codespell` and update `dependabot`

SauravMaheshkar:
Changes proposed by this PR can be summarized as follows :-

* Move `codespell` config to `.github/config` dir for a simpler project structure and thereby placing all configurations within the same folder. A corresponding change was made to the `pre-commit` configuration.
* Addition of `bundler` to the `dependabot` configuration file.


SauravMaheshkar:
@klieret Request for Review

klieret:
Since the deployed version of this webpage is built from github pages, wouldn't we already get reasonably new versions from all the Gems? 

The dependabot would basically replace lower bound with exact versions and then keep them updated, right?

So probably the only clear effect of this would be to force people to use the same local versions of the Gems that github pages uses. 

What do you think @hegner ?

As for the `codespell.txt`, I'm a bit conflicted. I also don't like crowded root directories, but `.github` doesn't seem like the right place for it (since it's not technically related to gh). So I'd prefer to leave it as it is.

hegner:
I think the `dependabot` one is OK.

For moving the `codespell.txt` I agree with @klieret. It is not GitHub specific. If we ever want to do some cleanup we would need to create a separate subdir for such checking tools and checking settings. Since one wants to use that locally too.

SauravMaheshkar:
Addressed 😄

PR #1401, opened 2023-03-21
Add redirections for HSF-hosted training material

klieret:
~~Probably the link checker will fail this...~~

Adding @hegner for review because we claim the `/t/` folder... (the domain name is already so long, didn't want to spell out training 😅 )

### DIFF
--- a/_data/training-modules.yaml
+++ b/_data/training-modules.yaml
@@ -21,7 +21,7 @@
   id: python
 - description: 'A full introduction to C++ based on a series of slides and exercises.'
   name: HEP C++ Course
-  repository: https://github.com/hsf-training/cpluspluscourse
+  repository: https://hepsoftwarefoundation.org/t/cpp
   status: stable
   videos: https://github.com/hsf-training/cpluspluscourse#-video-recordings--events
   webpage: https://github.com/hsf-training/cpluspluscourse#-getting-the-latest-pdf
@@ -31,7 +31,7 @@
   repository: https://github.com/hsf-training/hsf-training-cmake-webpage
   status: stable
   videos: ''
-  webpage: https://hsf-training.github.io/hsf-training-cmake-webpage/
+  webpage: https://hepsoftwarefoundation.org/t/cmake
   id: cmake
 - description: ''
   name: Distributed file systems and grid computing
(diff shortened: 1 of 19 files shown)


klieret:
(there are some follow up discussions with Wouter on slack)

hegner (changes requested):
The "t" is fine with me. If we want to have other kind of permalinks, maybe "s" for short or "p" for permanent could be a better option in the long rum

hegner:
@klieret - ping

klieret:
I was thinking about `/t` for training (so we don't need to attach a `-training` suffix or so), though I also wouldn't be opposed to doing `/s` or `/p`.

@wouter Was thinking about just taking `/training`, but I'm a bit against that just because it would flood the `/training` folder with these silly files (wish there was a better solution to perhaps just configure all redirects with a single yaml file, but I didn't immediately find one)

I'm currently travelling, I'm available for more discussion toward end of next week :)

klieret:
Hi @hegner : I talked with Wouter and he's ok with `/t/` (for 'training') as originally proposed. If you are as well, then we can merge this :)   If not, let's discuss :)

klieret:
Hi @hegner , ready to review?

hegner:
@kilian - two things:
1) you should document it
2) I do not like commits of auto-generated files here. There is no way in automating it for GitHub pages?

klieret:
1. Will do
2. I searched but did not find anything jekyll-based 🤷. We could however probably do it with github actions. For that, we'd first have to make the github pages step explicit (i.e., using [this action](https://github.com/marketplace/actions/deploy-to-github-pages), for example), and add another step. I'd personally say that that's not quite worth the increase in complexity, but I'd be open to implement it...

hegner:
OK. Then we ignore point 2 for the moment. Given documentation...

hegner (approved):

klieret:
I've done some more thinking and I think one of the issues will be that most people will probably still copy the URL from the address bar after opening the page (which will then be that of the github pages). I have to do some reading on whether and how we could achieve something better...

hegner:
OK. So could you then make this [WIP] ? So that we know that this isn't yet what we want to do? Thanks.

klieret:
Thanks for the suggestion, converted to draft.

hegner:
@klieret - do you still want to work in this direction or do we close this ticket?

klieret:
I currently don't have the resources to work on this. So let's close it perhaps for now.

PR #1405, opened 2023-03-24
add cvmfs gsdocs proposal

vvolkl:


hegner:
@vvolkl  - you would need to update _activities/gsdocs.md and preserve the information contained in there in a 2022 document. Have a look at #1404 for the technical details

PR #1414, opened 2023-05-02
Add carpentry event

klieret:


wdconinc (approved):
The mobile github interface is really quite pointless when showing this diff... It strips the leading spaces and the +/- from the diff start to bleed into the - from the yaml...

PR #1415, opened 2023-05-22
Add Plausible web analysis to curriculum page

klieret:


klieret:
Hi @wdconinc , can you review?

wdconinc (approved):
Looks fine to me. I can't actually test it locally anymore since github pages is still stuck on ruby 2 and the rest of the world seems to have moved on to 3. But the change seems harmless.

wdconinc:
Let's just assume we don't violate GDPR with this.

PR #1424, opened 2023-06-28
Update training-schools.yml

fatala22:
XIV edition of the ESC School


klieret (approved):
Thanks a lot for adding this @fatala22 !

PR #1425, opened 2023-06-29
Add schools

klieret:


wdconinc (approved):

wdconinc:
Do we already have codas-hep that we can add too?

klieret:
>Do we already have codas-hep that we can add too?

codas-hep is already added

PR #1429, opened 2023-07-06
fix: rm twitter inset since non-operational

wdconinc:
This removes the twitter inset that doesn't show anything useful due to twitter policy changes (see below). We could replace it with a link to the twitter profile and other social profiles, but I don't think that would necessarily fit with the layout here.

image

### DIFF
--- a/_layouts/main.html
+++ b/_layouts/main.html
@@ -96,15 +96,3 @@ in High Energy Physics software and computing internationally.
 
   </div>
 </div>
-
-<div class="row" style="margin-top: 30px">
-
-  <div class="col-lg-3"></div>
-
-  <div class="col-lg-6">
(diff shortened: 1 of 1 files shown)


graeme-a-stewart (approved):

graeme-a-stewart:
Thanks @wdconinc - let's just disable it for now. As we are rethinking the website we can see where social media links best go in the future.

klieret:
Oh wow, that's too bad. Do you have any link for the policy changes, I wonder what this is about?

wdconinc:
https://www.theverge.com/2023/6/30/23779764/twitter-blocks-unregistered-users-account-tweets for example

klieret:
thanks. This is probably something slightly different (because I can't see the tweets despite being logged in), but this really impacts the usefulness...

PR #1434, opened 2023-07-13
Add roy as educator

klieret:


_profiles/roy_frank_cruz.md
```diff
@@ -0,0 +1,25 @@
+---
+# Required:
+title: "Roy Frank Cruz"     # first (middle) last
+country: US              # your country of residence (2 capital letters, e.g. US, GB, DE)
+
+# Optional (please add github or gravatar for your picture)
+github:     55238184                          # your github ID (not full url)
```
wdconinc, line 7: Is this github id right? Same for gitlab id and linkedin id. Is this may the numerical id below instead of the handle?
image
In any case, the link to github.com/<number> won't work.

klieret: Thanks! I've reached out to Roy to clarify

klieret: @wdconinc Ready for review

PR #1438, opened 2023-08-13
Add prettier pre-commit hook

klieret:
This hook provides standardized formatting for a variety of files, including markdown (where it mostly wraps text) and yaml (I've removed HTML and CSS because it struggled with some custom elements there).

I'd say we want at least _some_ of this, but this also now changes almost every single file, meaning that hanging PRs will most likely yield merge conflicts (that can be resolved by applying this hook there as well).

I'm opening this PR to discuss about this ;) 

EDIT: Something still seems to go wrong when applying this in CI


eduardo-rodrigues:
Hello @graeme-a-stewart, @hegner, @jouvin, I would be tempted to close this PR at this point. It is massive and with ideas of "upgrading" the site, it seems less relevant.

graeme-a-stewart:
Hi @eduardo-rodrigues - I would agree, given the medium term plan to change website technology, a large formatting PR isn't so useful, so let's close this one.

PR #1441, opened 2023-08-13
Add training modules from other initiatives

klieret:
The added modules are of somewhat different format, but they cover blindspots that the HSF modules have. 

Also it marks the reana training as stable (we've had a student walk through all of them and he didn't have issues)

### DIFF
--- a/_data/training-modules.yaml
+++ b/_data/training-modules.yaml
@@ -1,3 +1,10 @@
+- description: 'This course covers various best practices like testing, pytest, object oriented programming, packing, CI, and more.'
+  name: "Software Engineering for Scientific Computing"
+  repository: https://github.com/henryiii/se-for-sci
+  status: stable
+  videos: ''
+  webpage: https://henryiii.github.io/se-for-sci/
+  id: se-for-sci
 - description: 'Track code changes, undo mistakes, collaborate. This module is a must.'
   name: Version controlling with git
   repository: https://github.com/swcarpentry/git-novice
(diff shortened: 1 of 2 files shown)


_data/training-modules.yaml
```diff
@@ -1,3 +1,17 @@
...
+  videos: ''
+  status: beta
+- description: 'This course covers various best practices like testing, pytest, object oriented programming, packing, CI, and more.'
+  name: "Software Engineering for Scientific Computing"
+  repository: https://github.com/henryiii/se-for-sci
+  status: stable
+  videos: ''
+  webpage: https://henryiii.github.io/se-for-sci/
+  id: se-for-sci
 - description: 'Track code changes, undo mistakes, collaborate. This module is a must.'
```
wdconinc, line 15:
```suggestion
 - description: 'Track code changes, undo mistakes, collaborate. This module is a must.'
```
This is more than a module and overlaps with many of our other modules. I'd avoid calling it a 'must' unless you specify something like 'for project maintainers' or so.

wdconinc (approved):

_data/training-modules.yaml
klieret: That's the description for the next one :)

klieret: (I think a yaml formatter originally sorted the keys alphabetically)

wdconinc: Haha, I see now.

PR #1444, opened 2023-08-14
Update training conveners

klieret:


wdconinc (approved):
Approving.

Twitter is dead, says the link checker. Marking with an ✖️.

PR #1448, opened 2023-08-14
Remove twitter references

klieret:
Since the link to the ML appears below, I removed the whole "News" block. 

We should find a replacement for Twitter as a microblog. Can't we just use Mastodon if we mostly want to use it as a newsfeed rather than reaching the world?

### DIFF
--- a/_workinggroups/training.md
+++ b/_workinggroups/training.md
@@ -28,11 +28,6 @@ The group aims to develop a training program that can be pursued by researchers 
 </div>
 
 
-## News
-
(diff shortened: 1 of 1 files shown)


wdconinc (commented):
Why not keep the mailing list?

klieret:
> Why not keep the mailing list?

Because there's a dedicated section below:

image

Because the mailing list is also not so much "news" and I don't think it makes sense to have a section with only one link in it 🤷

wdconinc (approved):

PR #1453, opened 2023-09-13
Add training center version for embedding as iframe

klieret:
As suggested by @henryiii during the IRIS-HEP retreat, it might be nice to embed the training center at the pages of our partners. For this, this PR creates a "blank page" with only the modules present (no HSF header/footer), no other content. One target would be IRIS-HEP or INTERSECT sites.

This required some refactoring to avoid duplication.

### DIFF
--- a/_includes/training_center.html
+++ b/_includes/training_center.html
@@ -1,0 +1,39 @@
+### Basics
+
+{% include list_of_selected_training_modules.html ids="unix,git,python,ssh,ml,matplotlib,root" %}
+
+### Software Development and Deployment
+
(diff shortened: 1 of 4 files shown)


klieret:
Link checker failure is expected

wdconinc:
Do you have this published on a fork somewhere?

klieret:
I tested it locally.

klieret:
Ah nvm, this works: https://klieret.github.io/hsf.github.io/training/center_iframe.html

wdconinc (approved):

klieret:
Ah, what I didn't realize is that opening links from an iframe is not supported in firefox due to some security policies (seems to work fine in chromium)...

PR #1463, opened 2023-11-09
Add Julia training modules in new Julia corner

klieret:
See discussion in https://github.com/JuliaHEP/JuliaHEP-2023/issues/31


_data/training-modules.yaml
```diff
@@ -177,3 +177,10 @@
   videos: ''
   webpage: https://hsf-training.github.io/hsf-training-reana-webpage/
   repository: https://github.com/hsf-training/hsf-training-reana-webpage
+- description: A Julia for HEP introduction especially for those familiar with Python (or C++)
```
wdconinc, line 180:
```suggestion
- description: An introduction to Julia for HEP, especially for those familiar with Python (or C++)
```

PR #1476, opened 2023-12-20
Insufficient bottom margin #1472

Hrittik20:

PR #1480, opened 2024-01-03
Increased bottom margin for the footer

### LINKED ISSUE #1472: Insufficient bottom margin
There is very little bottom margin for the line in footer, which is present in every page, except the homepage. It must be increased a bit.

Screenshot 2023-12-13 at 10 18 56 PM

gunjjoshi:
Added a new class named "user-footer" in the footer inside default.html.
Set a bottom margin for the same class in hsf.css

Before:
Screenshot 2024-01-03 at 11 22 47

After:
Screenshot 2024-01-03 at 11 23 21

Now, it looks better.

Closes #1472

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,39 +1,55 @@
 <!DOCTYPE html>
 <html>
-<head lang="en">
-    <meta charset="UTF-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1 user-scalable=no">
+  <head lang="en">
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta
+      name="viewport"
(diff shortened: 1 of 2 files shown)


klieret:
Hi @gunjjoshi . Thank you for your PR. But wasn't this addressed in #1476 ?

klieret:
I'm going to close this, because I believe it's a duplicate of the earlier #1476 , but let me know if this is incorrect.

gunjjoshi:
> Hi @gunjjoshi . Thank you for your PR. But wasn't this addressed in #1476 ?
> 
> Also I see a lot of unrelated formatting (?) changes.

There is an issue over there, as mentioned by you:
image

So, for this reason, I made another PR, fixing the same thing, now the margin is only from the bottom, unlike the one in #1476 , which is coming both upwards and downwards.

klieret:
#1476 was updated in the meantime to only affect the bottom though (and this is the version which I just merged and that is now live at https://hepsoftwarefoundation.org/)

PR #1484, opened 2024-01-22
Add link for workshop

graeme-a-stewart:
Add link to workshop Indico


graeme-a-stewart:
Trivial update, so self-merging as tests pass

PR #1490, opened 2024-02-04
add_profile

slackuj:


vvolkl:
Hi Umesh,

I see you've just written me an application to GSoC. The profiles here are intended for members of the HSF training group. There's no need to add yourself here, so I'm closing this PR.

PR #1492, opened 2024-02-05
Added Ganga GSoc 2024 poposal

egede:
A proposal for a GSoC project from the Ganga team. As encouraged by Google the project is AI related.


_gsocproposals/2024/proposal_GangaAIassistant.md
```diff
@@ -0,0 +1,44 @@
...
+
+As a scripting and command line interface, there will naturally be usrs that have problems with getting the syntax correct. To solve this, they will often spend time searching through mailing lists, FAQs and discussion fora or indeed just wait for another more advanced coder to debug their problem. The idea of this project is to integrate a Large Language Model (LLM) into the command prompt in Ganga. This should allow the user to describe in words what they would like to do and get an example that they can incorporate. It should also intercept exceptions thrown by the Ganga interface, help the user to understand them and propose solutions.
```
vvolkl, line 17:
```suggestion
As a scripting and command line interface, there will naturally be users that have problems with getting the syntax correct. To solve this, they will often spend time searching through mailing lists, FAQs and discussion fora or indeed just wait for another more advanced coder to debug their problem. The idea of this project is to integrate a Large Language Model (LLM) into the command prompt in Ganga. This should allow the user to describe in words what they would like to do and get an example that they can incorporate. It should also intercept exceptions thrown by the Ganga interface, help the user to understand them and propose solutions.
```

vvolkl (approved):
Thanks for the submission!

In general, AI/ML themed tend to attract many applications. Maybe it's wise to suggest an email subject for candidates, to make it easier to create inbox filters?

PR #1503, opened 2024-02-07
[gsoc24] remove repetition of Cppyy project name

maximusron:
This PR updates the title of a Cppyy proposal to prevent repetition since the project name is now prepended to the title


vvolkl (approved):
Thanks!

PR #1527, opened 2024-02-29
Fix whitespaces in ANL description

maszyman:


vvolkl (changes requested):
The parsing here is a bit confusing: To make sure the whitespace shows up, you need to remove all the linebreaks in `descriptions`. We need to fix this at one point to allow multilines.

PR #1534, opened 2024-03-11
Close submission of evaluations for Julia ML project

graeme-a-stewart:
We have more than enough students now, so we'd rather not have more solutions submitted


wdconinc (approved):

graeme-a-stewart:
Thanks @wdconinc!

Issue #1512, opened 2024-02-12
The markdown-to-HTML process did not pick up 2 of the 4 HSF Training conveners


jpivarski:
The PR is merged, but the HSF Training Working Group page looks like this:

![image](https://github.com/HSF/hsf.github.io/assets/1852447/8c85655e-78f2-477d-9e8a-0a60830b282c)

The HTML only has the two of us, too:

```html
<div class="row">


<!-- to do delayed loading etc., see https://www.ratanparai.com/jekyll/Responsive-image-on-jekyll/ and then the original carpentries template -->

<div class="medium-3 columns">
<div class="team-member anchor-offset" id="amorenobr">
<a href="/profiles/alexander_moreno.html"><!-- Make sure to include educator_header to use this -->


<img data-src="https://avatars.githubusercontent.com/amorenobr" class="img-circle lazyload" alt="GitHub profile photo of Alexander Moreno Briceño">

</a>
<h3><a href="/profiles/alexander_moreno.html">Alexander Moreno Briceño</a></h3>
<!-- social badges for people -->
<!-- this needs the font awesome css to be included, e.g. by including profile_header.html -->

<ul class="list-inline social-buttons">

<li> <a href="https://github.com/amorenobr"> <i class="fab fa-github"></i> </a> </li> 
<li> <a href="https://gitlab.com/amorenobr"> <i class="fab fa-gitlab"></i> </a> </li> 

<li> <a href="https://orcid.org/0000-0001-8415-2543"> <i class="fab fa-orcid"></i> </a> </li> 
<li> <a href="https://www.linkedin.com/in/alexander-moreno-brice%C3%B1o-a4b80924a"> <i class="fab fa-linkedin"></i> </a> </li> 
<li> <a href="mailto:alexander.moreno@uan.edu.co"> <i class="fa fa-envelope"></i> </a> </li> 

</ul>

</div>
</div>

<!-- big long gap of empty lines -->

<!-- to do delayed loading etc., see https://www.ratanparai.com/jekyll/Responsive-image-on-jekyll/ and then the original carpentries template -->

<div class="medium-3 columns">
<div class="team-member anchor-offset" id="jpivarski">
<a href="/profiles/jim_pivarski.html"><!-- Make sure to include educator_header to use this -->


<img data-src="https://avatars.githubusercontent.com/jpivarski" class="img-circle lazyload" alt="GitHub profile photo of Jim Pivarski">

</a>
[...]

PR #1537, opened 2024-03-18
fix: conveners list on HSF Training

jpivarski:
I was about to give up on #1512 and replace the fancy list with a simple bulleted list, but then I noticed that Lera's and Holly's names were not exactly the same as the `title` attribute in their profiles. So I'll try that first, and if it renders correctly, #1512 is fixed. If not, we use a bulleted list.

### DIFF
--- a/_workinggroups/training.md
+++ b/_workinggroups/training.md
@@ -141,7 +141,7 @@ We are always looking for volunteers from the community to help us with our trai
 {% endif %}
 ## Conveners
 
-{% assign persons = "Alexander Moreno Briceño, Valeriia Lukashenko, Jim Pivarski, Holly Szumila" | split: ", " %}
+{% assign persons = "Alexander Moreno Briceño, Lera Lukashenko, Jim Pivarski, Holly Szumila-Vance" | split: ", " %}
 {% include list_of_selected_profiles.html %}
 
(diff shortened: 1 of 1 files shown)


hegner (approved):

jpivarski:
It didn't work:

image

So I'm going to open another PR to replace the panel of images with a simple bulleted list.

hegner:
There must be a bug in their profiles. I would have expected them to appear here:
  https://hepsoftwarefoundation.org/training/community.html

no?

hegner:
Yes. they need to be changed to YAML instead of config style

jpivarski:
The ones that work:

* https://raw.githubusercontent.com/HSF/hsf.github.io/main/_profiles/alexander_moreno.md
* https://raw.githubusercontent.com/HSF/hsf.github.io/main/_profiles/jim_pivarski.md

The ones that don't work:

* https://raw.githubusercontent.com/HSF/hsf.github.io/main/_profiles/lera_lukashenko.md
* https://raw.githubusercontent.com/HSF/hsf.github.io/main/_profiles/holly_szumila.md

Is it because the ones that don't work are missing a `---` at the top of the file? They all have a `---` at the bottom of the file (neglecting comments).

hegner:
yeah. that could be indeed

hegner:
@jpivarski problem solved! :-)

PR #1539, opened 2024-03-20
fix: profiles need to be formatted differently

jpivarski:
Following up on #1537.


hegner:
@jpivarski - had to close/reopen due to a problem in the infrastructure

PR #1553, opened 2024-05-06
Redirect to new training center

ariostas:
Now that the new training center is live (https://hsf-training.org), we should redirect the old page to the new domain. For now, I didn't remove anything in case we need to revert back if there are issues with the new one, but it would be good to have the new one fully deployed and get some feedback on what could be improved.


jpivarski (approved):
:+1:

jpivarski:
(We approved this in our regular meeting.)

PR #1557, opened 2024-06-14
adding info about the steering group meetings

roiser:
adding info about the steering group meetings

please check, I've added also the indico category at the bottom of the file. The name of the category is still "Coordination/Organisation Meetings", maybe change it to "Steering Group Meetings"?


hegner:
Thanks! I am not sure whether the better place isn't in the `about` section, where we as well describe the coordination team, which is changed now as well. I think we should update the two pages in sync.

roiser:
Thanks, I've moved it into the About section now

graeme-a-stewart (approved):

PR #1561, opened 2024-07-02
Change centre column to JuliaHEP 2024

graeme-a-stewart:
JuliaHEP 2024 registration and abstract submission are open now, so make this prominent!


peremato (approved):
Thanks!

PR #1570, opened 2024-07-24
Not running Frameworks as a WG

graeme-a-stewart:
Remove WG from the list


graeme-a-stewart:
The downside of this change is that all memory of the Frameworks group disappears from the website

graeme-a-stewart:
OK, I have a new plan here where we can have archived activities (it's not just frameworks, there are quite a few), so this PR isn't relevant anymore.

PR #1579, opened 2024-08-29
Update documentation about editing website

jouvin:
- CodiMD section updated to reflect that the Hackmd.io service is not really an option anymore and suggest CERNBox for those who can use it
- repetition removed in running_meetings document

### DIFF
--- a/howto-website.md
+++ b/howto-website.md
@@ -6,7 +6,7 @@ layout: default
 
 ## About the HSF website
 
-This site is maintained by the HSF GitHub [contributors](https://github.com/orgs/HSF/people). If you're interested to become one contact the [HSF Steering Group]({{ site.baseurl }}/organization/team.html) or any team member. It was set up by Torre Wenaus and Benedikt Hegner.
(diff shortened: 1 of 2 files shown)


jouvin:
I have question: is it intentional that we don't enable automatic heading anchors? I find them convenient to allow referencing a specific section in a page but may be there was good reason not to enable them. How to enable them seems theme dependent...

graeme-a-stewart (approved):
LGTM!

graeme-a-stewart:
> I have question: is it intentional that we don't enable automatic heading anchors? I find them convenient to allow referencing a specific section in a page but may be there was good reason not to enable them. How to enable them seems theme dependent...

I would be in favour of switching these on if there was an easy way to do it (I have no sense of why wasn't done - was it difficult in the past?).

At least on the pages with a ToC its easy to get the heading anchor link, but on other pages far less so...

hegner:
We did not have it in the past just for the reason nobody asked for it.

PR #1609, opened 2024-10-28
fix codespell

vvolkl:
Add github username of gsoc contributor to whitelist. I would have thought that there's another way to avoid codespell to trigger as this only appears  in URLs but haven't found anything. This introduces the risk of future misspellings of "extinct", but I'm  just going to hope that nothing goes extinct and the risk remains small :)

### DIFF
--- a/codespell.txt
+++ b/codespell.txt
@@ -19,3 +19,4 @@ caf
 gage
 gauge
 re-use
+extint


graeme-a-stewart:
Thanks @vvolkl - yes, this one was becoming annoying and I think the benefits very much outweigh the risks!

graeme-a-stewart (approved):

PR #1619, opened 2024-11-20
Update seminar organisers

graeme-a-stewart:


graeme-a-stewart:
Pretty trivial, so self-approving

PR #1620, opened 2024-11-21
Add cached versions of badges

graeme-a-stewart:
When adding the HSF project badge to [prmon](https://github.com/HSF/prmon) I could not get the badge to work from a direct URL query to shields.io. Not clear why that happened as page previews of the README file worked fine.

This PR adds direct links to cached versions of the badges, which does seem to work correctly.

N.B. There will be a link checker failure, because referenced files do not exist until this PR is accepted.


hegner:
Looks like a better solution

hegner (approved):

eduardo-rodrigues (approved):
Great that you managed to sort the "funny" issue so quickly :+1: !

PR #1630, opened 2025-01-16
[gsoc] 2024->2025

vvolkl:


eduardo-rodrigues:
Hi. One quick question from me - do the 2024 pages with projects, mentors, etc., get archived somewhere reachable or simply get hidden away on the GitHub repo? Having the ability to find out about the past projects and great achievements seems important to me. Thanks.

vvolkl:
That's a very good point, indeed it would be intention that pages from previous years are archived under the url

https://hepsoftwarefoundation.org/gsoc/2023/index.html

which are linked to here: https://github.com/HSF/hsf.github.io/blob/main/_activities/gsoc.md?plain=1#L27

I did forget to update that for 2024 though, so thanks for bringing that up!

graeme-a-stewart (approved):
Looks good to me - thanks Valentin!

PR #1633, opened 2025-01-23
Add two MCnet projects

agbuckley:
Glasgow and UCL orgs and MCnet project reinstated for 2025; two new project proposals from MCnet.


vvolkl (approved):
Great projects; thanks!

PR #1640, opened 2025-01-30
Clarify info about the GSoC proposal submission

maszyman:
Adding small bullet point emphasising the need for submitting the proposals via GSoC system, not only to mentors (it was not clear in the past for my 2 students who did not do it and could not be later selected).


vvolkl (approved):
Very true, thanks!

PR #1685, opened 2025-02-11
Add GSoC proposal for ATLAS lossy float compression study

maszyman:
cc @gemmeren


vvolkl (approved):
Great project, thanks!

Issue #1696, opened 2025-02-13
gsoc: add mentors to metadata, generate mentor list automatically


vvolkl:
It'd be nice if we could fill the page https://hepsoftwarefoundation.org/gsoc/2025/mentors.html automatically with the information from the project proposals https://github.com/HSF/hsf.github.io/tree/main/_gsocproposals/2025 . Probably the easiest is to move the mentor information in the yaml header, similar to what is done with the project length/duration

MohanadKh03:
Hello I have come up with a solution but there is a missing case that I am not sure how to handle
Some mentors do not have an organization name in the proposals so should they are added automatically to the mentors.html they would not have an organization name is that okay or how should it be handled ?
example: 

![Image](https://github.com/user-attachments/assets/7d1645a0-6103-43c2-bb95-d7f4eb156d6c)

vvolkl:
Nice! The organization should in principle be mentioned in the proposal, it can just be missing in the list if not.

There should be no duplicated entries though.

PR #1700, opened 2025-02-13
Issue #1696: Add mentors to metadata, generate mentor list automatically

MohanadKh03:
PR Summary: 
- Mentor names are extracted automatically from the GSoC proposal pages from their front matter "YAML format"
- The mentors are listed uniquely by their email on the mentors page

### DIFF
--- a/.github/config/mdcheck.json
+++ b/.github/config/mdcheck.json
@@ -1,6 +1,9 @@
 {
   "ignorePatterns": [
     {
+      "pattern": "https://mybinder.org"
+    },
+    {
       "pattern": "https://www.sciencedirect.com/science/article/abs/pii*"
     },
     {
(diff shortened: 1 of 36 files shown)


vvolkl:
This is fantastic, thank you very much for this contribution!

One small comment: names are unfortunately a bit complicated to handle, and here we have one case where the mentor has a last name and first name with whitespace. (First name: "John S." Last name: "De Stefano Jr." 
I'd suggest to just add a firstname and lastname field to the yaml header, like that you don't need to split the full name.

MohanadKh03:
> I'd suggest to just add a firstname and lastname field to the yaml header, like that you don't need to split the full name.

Done

My bad I made a mistake while resolving the conflict but now it is fixed

vvolkl:
I've added some small fixes, now this LGTM:

 - mentor section in proposals from 2025 onwards. old
  projects do not have the metadata.
- add is_preferred_contact that just prints the mentor in bold
- change mentor.role to mentor.additional_information for more
  flexibility

Your work was very easy to extend, thanks again for this nice contribution!

vvolkl (approved):

vvolkl:
Fixes #1696

PR #1702, opened 2025-02-17
Add missing emails to GEANT4 GSoC proposal

aaronj0:
I noticed missing emails after the PR automating mentor info. I'd like to confirm with @Piyush-555 if the added emails are correct.


Piyush-555:
@aaronj0 Yes, the emails are correct. However, I'd like to avoid those emails added there, unless it breaks other things. It gets a bit hard to keep track of emails when candidates are inconsistent with mailing just the primary supervisor.

But if it's necessary you can add the emails, they are correct.

vvolkl:
I think that makes sense, you do have one point of contact listed, and keeping the adress off the page will indeed help with candidates writing to the preferred contact

PR #1716, opened 2025-03-11
gsoc: add summaries for 2011 and 2012

vvolkl:
There are some rather nice summary reports from the first editions of GSoC on the EP-SFT website (https://ep-dep-sft.web.cern.ch/project/cern-sft-google-summer-code). Even though this was not yet organized by the HSF I though it makes sense to have those all on the same website.

@graeme-a-stewart Let me know if there are any objections to this


graeme-a-stewart (approved):
Nice idea @vvolkl - thanks for digging out this history!

PR #1728, opened 2025-04-21
Use direct links to training center

ariostas:
This PR changes the links to the training center so that they point directly there instead of getting there through a series of redirections. The Plausible Analytics for the training center says that most people are getting there with a direct link, but I suspect that the reason for this is that the redirections are obscuring the real sources. So we will see if after making this change we see that most people are really coming from the HSF website.


ariostas:
I fixed a link that was broken, but the other two broken links reported by the `Check Markdown links` workflow are false positives. They probably just rate-limited GitHub's servers.

ariostas:
@michmx I can't request a reviewer, but could you take a look at this when you get a chance?

michmx (approved):
Thanks for the fixes!

PR #1735, opened 2025-06-04
List AG membership

graeme-a-stewart:
It seems we overlooked actually listing the members of the Advisory Group


eduardo-rodrigues (approved):
Hi Graeme, this is fine, of course. But given that we have the "Composition" section a bit above, this addition could be made a subsection "Current representatives composition" (or a similar title) underneath it. WDYT?

PR #1747, opened 2025-07-09
Replace Belle II with CMS.

chrishanw:
I moved from Belle II to CMS at DESY, though I will still remain a bit active in Belle II for the next few months, but I guess it's OK not to mention Belle II any longer.


hegner (approved):
Thanks for the update. Have fun in CMS!

chrishanw:
> Thanks for the update. Have fun in CMS!

Thanks a lot!

PR #1750, opened 2025-08-01
Added (brief) 2025-07-31-coordination minutes

urania277:
Meeting was cancelled, this contains a record of the discussion that we had on seminar slots after the vacation period.


cantel (approved):
I just updated the date. Looks good to me now!

PR #1754, opened 2025-08-27
Fix Peter's affiliation

maszyman:
https://github.com/HSF/hsf.github.io/commit/9b4eb73012b1e79048aaba2f4d22627b086935b6 introduced wrong affiliation.

@MohanadKh03 @vvolkl


eduardo-rodrigues (approved):
Thanks for the fix. I will merge right now since trivial

Issue #1766, opened 2025-10-14
Improve Navbar Usability by Enabling Hover-to-Open Menus


this-is-varunxyz:
Is your feature request related to a problem?
Currently, the navigation bar dropdown menus on the website require a user to click to open them. This adds an extra step to navigation and can feel less intuitive than modern web standards where menus often open on hover. For users who are exploring the site, this can make the experience slightly less fluid.

Describe the solution you'd like
I propose that the main navigation bar's dropdown menus be updated to open when a user hovers their mouse over a top-level menu item (e.g., "Meetings," "Communication"). The menu should remain open as long as the user's mouse is over the menu item or the dropdown menu itself. The menu should close when the user's mouse moves away.

This change would make the navigation feel more responsive and align with the user experience of many other websites.

Additional context
This is a relatively small user experience improvement that can make a positive impact on how users interact with the site. Thank you for your consideration!

<img width="1437" height="362" alt="Image" src="https://github.com/user-attachments/assets/44e9b688-2196-45a1-8b1a-236f643f4992" />

eduardo-rodrigues:
Closed via https://github.com/HSF/hsf.github.io/pull/1768.

PR #1768, opened 2025-10-16
Feat: Implement hover-to-open navbar with animation

this-is-varunxyz:
This PR implements a hover-to-open feature for the main navigation bar on desktop-sized screens to improve usability. #1766

**Key improvements:**
- Menus now open on hover with a smooth slide-down animation.
- On desktop, the default click action on the main links is disabled to prevent confusion and remove the `:focus` state.
- The feature is disabled on mobile viewports (<768px), preserving the default tap-to-open behavior.

### DIFF
--- a/_includes/navbar.ext
+++ b/_includes/navbar.ext
@@ -26,11 +26,11 @@
             </li>
             <li class="dropdown">
               <a class="dropdown-toggle" data-toggle="dropdown" href="{{site.baseurl}}/index.html" id="meetings_menu"><span class="glyphicon glyphicon-phone-alt"></span>&nbsp;&nbsp;Meetings<span class="caret"></span></a>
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
Hi @this-is-varunxyz. This is looking very good. I checked the effect on a couple of browsers and my phone, and all goes as expected. I requested that a couple of other colleagues from the team look at this PR since it changes the feeling of the site a bit.

N.B.: if you took the code from somewhere, do add a comment with the source, to be fair.

Thank you for this :+1:!

this-is-varunxyz:
thank you sir , @eduardo-rodrigues 
Yes, the code is original. I did used Gemini to generate the comments to ensure that the logic is easy for others to understand.

PR #1825, opened 2026-01-30
Patch for case and max width

graeme-a-stewart:
There was a case mistake in the merged centre column patch (#1821), this fixes that (this was not evident on OS X, which is a case insensitive filesystem!).

Also, limit the maximum image size of the GSoC logo to 250px (helps on medium screen sizes).


graeme-a-stewart:
Thanks - fix is merged!

PR #1837, opened 2026-02-01
feat: insert CppCon link to PODIO GSoC proposal

veprbl:
Addressing suggestion https://github.com/HSF/hsf.github.io/pull/1832#pullrequestreview-3735279115. This is probably irrelevant to the proposal, but I'm happy to include, if people agree that it may be of interest to participants.


wdconinc (approved):
Agreed that it's a bit disconnected from the core of this project. But being name checked by Herb may indeed be useful to highlight and attract good candidates ;-)

PR #1847, opened 2026-02-03
gsoc 26: fix metadata for GreenMetaData org / BOA org

maszyman:


caterina-doglioni:
Thanks, something strange happened with the checks yesterday so I was going to see what would happen today. I'll double check and get back to you when ready.

vvolkl (approved):
LGTM!

@caterina-doglioni just as formality, I checked the license of boa-constrictor, and while zenodo says it is AGPL, there is no LICENSE file in the github repository. You may want to consider adding one.

PR #1852, opened 2026-02-05
Update training conveners and Andres Rios-Tascon profile

ariostas:
Just updated some of my info and added my name to the list of training conveners.


eduardo-rodrigues (approved):
Thank you for this.

PR #1928, opened 2026-05-14
DIRAC / DIRACX becomes an affiliated project

eduardo-rodrigues:
FYI @ruslan33.


eduardo-rodrigues:
(I'm merging myself since this has already been approved and the update is trivial.)

PR #1930, opened 2026-05-19
Update conveners of Software Dev. Tools and Packaging

eduardo-rodrigues:
FYI @ruslan33.  (While at it I reordered to have the most recent conveners at the top of the list, which I think makes more sense.)


eduardo-rodrigues:
Merging myself as the rendering is fine and this is a trivial update.

PR #1932, opened 2026-06-03
Final update of AA conveners for 2026

eduardo-rodrigues:


eduardo-rodrigues:
Merging myself this trivial (and long due!) update.

PR #1942, opened 2026-07-08
Remove obsolete and basically empty inventory page

eduardo-rodrigues:
Hi all. This page is in my opinion unused since a while and obsoleted by the affiliation programme. I'm hence suggesting to remove it.

It is only referenced (after this PR) in the minutes of the coordination meeting https://hepsoftwarefoundation.org/organization/2018/06/28/coordination.html but that seems old enough not to create an issue; it dates from the early days when the inventory was suggested.


sextonkennedy:
Looks fine to me.  I agree that the affiliation program supersedes this inventory.

sextonkennedy (approved):
Looks good

PR #1944, opened 2026-07-16
Be consistent with the word convener

eduardo-rodrigues:
Indeed there were a few "convenor" left over here and there.

Note that I did not change the minutes as they are what they are. But nicer to have consistency among the website (active, as it were) pages.


eduardo-rodrigues:
Merging since a no-brainer.

PR #1956, opened 2026-08-17
Add 2026 to training years for Richa Sharma

richa2710:


michmx (approved):
A fake negative in the link checks:
```
  [✖] https://www.linkedin.com/in/richa-sharma-71606a72/ → Status: 999
```
this link is actually valid. Nothing to do.

PR #1957, opened 2026-08-17
Update statistics

richa2710:


_profiles/richa_sharma.md
```diff
@@ -4,6 +4,7 @@ training_years:
 - 2023
 - 2024
 - 2025
+- 2026
```
michmx, line 7: Duplicates https://github.com/HSF/hsf.github.io/pull/1956. Probably ok.

michmx: It triggers the false alert on the link that is valid, as mentioned in https://github.com/HSF/hsf.github.io/pull/1956

michmx (approved):

PR #1960, opened 2026-08-19
Add training events for 2025 and 2026

michmx:
It was a while since we updated the page with training events. I am adding the ones from 2025 and 2026, either organized by us or with our participation.


amorenobr (approved):

amorenobr:
Thanks a lot Michel!

